PR #1, opened 2020-06-22
Create boilerplate code

laptou:
- Create directory structure with empty CSS, JS, and HTML files
- Add configuration for linters
- Add GitHub CI to enforce linting


geekster777:
You should request reviews from the other interns before requesting reviews from intern hosts.

LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Ibiyemi Abiodun
```
jessicaslaughter, line 3: This should be updated to reflect everyone in the group contributing to this project

client/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "lint:script": "eslint .",
+    "lint:style": "stylelint .",
+    "lint:fix:script": "eslint --fix .",
+    "lint:fix:style": "stylelint --fix .",
+    "lint": "run-s lint:script lint:style",
+    "lint:fix": "run-s lint:fix:*"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/laptou/step-capstone.git"
```
jessicaslaughter, line 30: This should be updated

client/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "contributors": [
+    "Ibiyemi Abiodun <ibiyemi@google.com>",
+    "Seun Omonije <seunomonije@google.com>",
+    "Grace Tian <gtian@google.com>"
+  ],
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/laptou/step-capstone/issues"
+  },
+  "homepage": "https://github.com/laptou/step-capstone#readme"
```
jessicaslaughter, line 41: Same

client/package-lock.json
```diff
@@ -0,0 +1,5879 @@
+{
```
geekster777, line 1: What consumes this file? I don't think it should be included in this repo. I expect it to change often and unnecessarily bloat PR's. Same with package.json (to a lesser degree). Please add it to the .gitignore.

client/.eslintignore
```diff
@@ -0,0 +1,3 @@
+node_modules/
```
geekster777, line 1: Why are there node_modules?

laptou: ESLint and Stylelint are Node.js programs. You need to install them via NPM in order to run them.

client/package-lock.json
laptou: This is consumed by the linting CI task. Like I said in my other project, `package-lock.json` stores _specific_ versions of packages that are installed, which is used by commands like `npm ci` to make sure that what is installed on another computer matches _exactly_ what is installed on your own, so that there are no mysterious errors. `package.json` only specifies approximate versions of packages, which means there is a chance that a package could get silently upgraded and break everything.

And it won't change often. Since I can't preprocess the frontend code, I can't use any Node.js dependencies at runtime anyway. This file and `package.json` are just for ESLint and Stylelint.

LICENSE
laptou: Sorry, that file was generated automatically and I forgot to update it.

client/.eslintignore
laptou: If you're worried about this making things complicated, don't be. I explained everything in the documentation I wrote in #3, and it's pretty straightforward.

client/package-lock.json
geekster777: My understanding was that linting would happen in the Github Workflow and potentially in a dev's code editor if they enabled it through an extension. We shouldn't be requiring everyone to download NodeJS just to enable linting. The burden of work is unnecessary and disproportionately affects the other team members. Both this and package.json should be removed.

laptou: If you don't want to put NodeJS on your computer (which is not hard, I wrote instructions on how to do it already), then you can just upload the PR and GitHub will lint it for you. If I don't have `package.json` in here, then GitHub cannot perform any linting.

client/.eslintignore
geekster777: Resolving since it looks like this helps you lint locally.

client/package-lock.json
geekster777: Fair enough. If it's needed for the github workflow, then it should be included here. Note, though, that teammates should not feel compelled to lint locally since Github will do it for them.

PR #10, opened 2020-06-25
Musixmatch backend

seunomonije:
### Backend framework for Musixmatch API:
Can't get the package I added to work. Decided to make my own.

#### Implementation notes:
 - Retrieves data from Musixmatch api and prints it raw on the screen.
 - Modular-ish design which will allow me to add more functions for more API calls. This is not the finished product.
 - MusixServlet.java contains hard-coded values for "Baby" by Justin Bieber. These values will later be retrieved by the front-end and sent to this servlet.
 - The two builder classes take care of the parsing away from the servlet to keep it clean.
 - MalformedUrlException is a reqiurement when using URL class. For now I'm not throwing errors but possibly will put try/catch inside functions. 

#### Results:
Screenshot 2020-06-24 at 5 16 09 PM


server/src/main/java/com/google/musicanalysis/site/ParametersBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+
+    switch (intendedUse) {
+      case "track.search":
+        this.trackName = params[0];
+        this.artistName = params[1];
+        break;
+      case "track.lyrics.get":
+        this.trackId = params[0];
+        break;
+    }
```
laptou, line 36: Add a default case. Probably throw an `IllegalArgumentException`.

server/src/main/java/com/google/musicanalysis/site/ParametersBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  /** Filters out the correct parameter string and returns it */
+  public String filter() {
```
laptou, line 40: This method name is unintuitive. You could rename it to `getParameters()`.

server/src/main/java/com/google/musicanalysis/site/RequestBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+   * @param operation the API method to send to Musixmatch
+   * @param parameters the parameter string to be sent to Musixmatch
+   */
+  public RequestBuilder(String operation, String parameters) {
+    this.operation = operation;
+    this.parameters = parameters;
+  }
+
+  /** @return the result from Musixmatch API */
+  public String grabResponse() throws MalformedURLException, IOException {
```
laptou, line 37: If you're just returning the string that is retrieved in `getResult()`, then why is this method needed?

server/src/main/java/com/google/musicanalysis/site/RequestBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+
+/*
+    Handles gathering data from the Musixmatch API.
+    This is structured the way it is becuase I will be adding more
+    to each function.
+*/
+public class RequestBuilder {
+
+  // Url building
+  private String baseUrl = "https://api.musixmatch.com/ws/1.1/";
```
laptou, line 17: These are constants that will _never_ change. Make them static and final, and name them in `SCREAMING_SNAKE_CASE`.

```suggestion
  private static final String BASE_URL = "https://api.musixmatch.com/ws/1.1/";
```

server/src/main/java/com/google/musicanalysis/site/ParametersBuilder.java
seunomonije: +1 Good idea.

server/src/main/java/com/google/musicanalysis/site/MusixServlet.java
```diff
@@ -0,0 +1,38 @@
...
+
+@WebServlet("/musix")
+public class MusixServlet extends HttpServlet {
+  @Override
+  protected void doGet(HttpServletRequest req, HttpServletResponse res)
+      throws ServletException, IOException {
+    res.getWriter().write("<h1>See below api data:\n\n</h1>");
+
+    // Setup variables, these will be given by the front end
+    String trackName = "Baby";
```
gracetian6, line 18: Thanks for these comments!

server/src/main/java/com/google/musicanalysis/site/RequestBuilder.java
seunomonije: I put this here to make it easiest on the readers as possible. When they scan through the code, they see that in order to get the response from the Musixmatch API, they instantiate a new class and run obj.grabResponse(), which will return as string. All of the work will remain in private functions to give an "under the hood effect". If they're interested in what's going on, they can look for more info, if not, they can just use it.

server/src/main/java/com/google/musicanalysis/site/RequestBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+   */
+  private URL buildUrl(String operation, String parameters) throws MalformedURLException {
+    String urlString = baseUrl + operation + outputReqs + parameters + apiKey;
+    URL url = new URL(urlString);
+    return url;
+  }
+
+  /**
+   * Calls the method and gets the result from Musixmatch. Reads data into an input buffer and
+   * returns as string.
```
gracetian6, line 56: kinda confused about why it needs to go through an input buffer, is this just to store the info before it gets called?

seunomonije: Ok, so what's basically happening here is when I call the API it basically returns to me a url to a site where the only contents on it is the response object. What I'm doing here is creating a buffer and reading the contents of that "site" line by line as a stream, and adding the contents to the buffer. 

I don't know the in's and outs about StringBuffer in Java, but to me, this is the equivalent to having an array in C with a buffer of _n_ chars, filling up that array with chars in an input stream, and then mallocing once you get close to the end to allocate more space. The buffer is necessary because we don't know how many characters the response is going to be. 

This link is where I got my info from: https://docs.oracle.com/javase/tutorial/networking/urls/readingURL.html

server/src/main/java/com/google/musicanalysis/site/MusixRequest.java
```diff
@@ -0,0 +1,75 @@
...
+   * @param operation the API method to send to Musixmatch
+   * @param parameters the parameter string to be sent to Musixmatch
+   */
+  public MusixRequest(String operation, String parameters) {
+    this.operation = operation;
+    this.parameters = parameters;
+  }
+
+  /** @return the result from Musixmatch API as string */
+  public String grabResponse() throws MalformedURLException, IOException {
```
laptou, line 35: Please explain why you are wrapping `getResult()` with `grabResponse()`. Are you planning to add more stuff to this method later?

server/src/main/java/com/google/musicanalysis/site/MusixParamBuilder.java
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Builds string to serve to request (track search)
+   *
+   * @param trackName name of the desired track
+   * @param artistName name of the desired artist
+   */
+  private String buildTrackString(String trackName, String artistName) {
+    // ready string for sending
+    try {
+      trackName = URLEncoder.encode(trackName, "UTF-8");
```
laptou, line 64: I discovered after making the earlier suggestion that you can use `StandardCharsets.UTF_8` instead of `"UTF-8"`, which will save you from needing the try-catch block. It's not a must-change, but if you want to simplify this.

You can import it like this:
```java
import java.nio.charset.StandardCharsets;
```

```suggestion
      trackName = URLEncoder.encode(trackName, StandardCharsets.UTF_8);
```

server/src/main/java/com/google/musicanalysis/site/MusixParamBuilder.java
```diff
@@ -0,0 +1,87 @@
...
+      artistName = URLEncoder.encode(artistName, "UTF-8");
+    } catch (UnsupportedEncodingException err) {
+      throw new AssertionError("Invalid URL");
+    }
```
laptou, line 68: I discovered after making the earlier suggestion that you can use `StandardCharsets.UTF_8` instead of `"UTF-8"`, which will save you from needing the try-catch block. It's not a must-change, but if you want to simplify this.

You can import it like this:
```java
import java.nio.charset.StandardCharsets;
```

```suggestion
    trackName = URLEncoder.encode(trackName, StandardCharsets.UTF_8);
    artistName = URLEncoder.encode(artistName, StandardCharsets.UTF_8);
```

server/src/main/java/com/google/musicanalysis/site/MusixParamBuilder.java
```diff
@@ -0,0 +1,87 @@
...
+      case "track.lyrics.get":
+        this.trackId = params[0];
+        break;
+      default:
+        throw new IllegalArgumentException("arguments required");
+    }
+  }
+
+  /** Filters out the correct parameter string and returns it */
+  public String filterParamString() {
```
laptou, line 45: Can I suggest `getFilterParamString()`?

server/src/main/java/com/google/musicanalysis/site/MusixParamBuilder.java
seunomonije: Sounds good to me

server/src/main/java/com/google/musicanalysis/site/MusixRequest.java
seunomonije: I did above. 
Copied:
I put this here to make it easiest on the readers as possible. When they scan through the code, they see that in order to get the response from the Musixmatch API, they instantiate a new class and run obj.grabResponse(), which will return as string. All of the work will remain in private functions to give an "under the hood effect". If they're interested in what's going on, they can look for more info, if not, they can just use it.

I am also planning to add more stuff later, as mentioned in the implementation notes and in the comments at the top of the file.

server/src/main/java/com/google/musicanalysis/site/MusixParamBuilder.java
seunomonije: grabFilteredParams() final offer.

server/src/main/java/com/google/musicanalysis/site/MusixParamBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+      case "track.lyrics.get":
+        this.trackId = params[0];
+        break;
+      default:
+        throw new IllegalArgumentException("arguments required");
+    }
+  }
+
+  /** Filters out the correct parameter string and returns it */
+  public String grabFilteredParams() {
```
geekster777, line 45: nit: name this getFilteredParams.

server/src/main/java/com/google/musicanalysis/site/MusixRequest.java
```diff
@@ -0,0 +1,75 @@
...
+    to each function.
+*/
+public class MusixRequest {
+
+  // Url building
+  private static final String BASE_URL = "https://api.musixmatch.com/ws/1.1/";
+  private String operation;
+  private static final String OUTPUT_REQS = "?format=jsonp&callback=callback&";
+  private String parameters;
+  private static final String apiKey = "&apikey=";
```
geekster777, line 21: nit: name this API_KEY since it's a const. Also, group all static variables together separately from your regular variables:

```
private String foo;
private String bar;

private static final ZOOP;
private static final DOOP;
```

server/src/main/java/com/google/musicanalysis/site/MusixRequest.java
geekster777: +1 to Ibiyemi. You should just make `getResult` public and remove `grabResponse`. Otherwise this obfuscation hurts readability more than improves it.

server/src/main/java/com/google/musicanalysis/site/MusixServlet.java
```diff
@@ -0,0 +1,38 @@
...
+    String artistName = "Justin Bieber";
+    String trackId = "32184842";
+    String use = "track.search";
```
geekster777, line 21: nit: name this `callType` or `requestType`

server/src/main/java/com/google/musicanalysis/site/MusixParamBuilder.java
jessicaslaughter: +1, "get" is the preferred verbiage, consistent with the concept of a "getter"

server/src/main/java/com/google/musicanalysis/site/MusixParamBuilder.java
```diff
@@ -0,0 +1,77 @@
+package com.google.musicanalysis.site;
+
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+
+/*
+    Builds parameter strings based off what the user wants.
+    This is structured the way it is becuase I will be adding more
+    to each function. I will clean it up later as well.
+*/
```
jessicaslaughter, line 10: nit: this should be formatted as javadoc, with asterisks before each line and two asterisks in the opening line

https://google.github.io/styleguide/javaguide.html#s7.1.1-javadoc-multi-line

server/src/main/java/com/google/musicanalysis/site/MusixRequest.java
```diff
@@ -0,0 +1,75 @@
+package com.google.musicanalysis.site;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.MalformedURLException;
+import java.net.URL;
+
+/*
```
jessicaslaughter, line 9: nit: same here, this should be formatted using javadoc style

server/src/main/java/com/google/musicanalysis/site/RequestBuilder.java
gracetian6: got it thank you

server/src/main/java/com/google/musicanalysis/site/MusixServlet.java
```diff
@@ -0,0 +1,46 @@
...
+    // Getting lyrics from a given id
+    res.getWriter().write("\n<h3>Getting lyrics from given id</h3>");
+    callType = "track.lyrics.get";
+    params = new MusixParamBuilder(callType, trackId);
+    request = new MusixRequest(callType, params.getFilteredParams());
+    res.getWriter().println(request.getResult());
+  }
+
+  /** @param user the user that will be converted to json */
+  private String convertToJsonUsingGson(String response) {
```
laptou, line 41: It doesn't seem like you're using this method any more (and I'm also not sure what it does). Please remove it if you're not using it.

server/src/main/java/com/google/musicanalysis/site/MusixRequest.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  /**
+   * Builds an appropriate URL with the information given.
+   *
+   * @param operation Musixmatch API method. See above.
+   * @param parameters The parameter string. See above.
+   */
+  private URL buildUrl(String operation, String parameters) throws MalformedURLException {
+    String urlString = BASE_URL + operation + OUTPUT_REQS + parameters + API_KEY;
```
laptou, line 41: Where are you retrieving the actual API key?

server/src/main/java/com/google/musicanalysis/site/MusixParamBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Builds parameter strings based off what the user wants. This is structured the way it is becuase
+ * I will be adding more to each function. I will clean it up later as well.
+ */
+public class MusixParamBuilder {
+
+  private String intendedUse;
```
geekster777, line 12: nit: name this requestType.

server/src/main/java/com/google/musicanalysis/site/MusixRequest.java
seunomonije: Atm just copy paste. Soon to be that other way that we mentioned but I have to look into it.

PR #11, opened 2020-06-25
Extended OAuth server

### LINKED ISSUE #2: Create server with OAuth capabilities
We need OAuth in order to be able to call the Spotify and YouTube APIs at all, so this is a high priority.

laptou:
- Extend OAuth server to also allow users to sign into YouTube
- Resolves #2

### DIFF
--- a/server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
+++ b/server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
@@ -32,9 +32,10 @@ public abstract class OAuthCallbackServlet extends HttpServlet {
    */
   protected abstract String getRedirectUri();
 
-  protected String getSessionServiceKey() {
-    return "oauth-state-" + getServiceName();
-  }
+  /**
+   * @return A key that is used to store authentication state in a session cookie.
+   */
+  protected abstract String getSessionServiceKey();
 
   @Override
   protected void doGet(HttpServletRequest req, HttpServletResponse res)
--- a/server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
+++ b/server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
@@ -32,9 +32,7 @@ public abstract class OAuthCallbackServlet extends HttpServlet {
    */
   protected abstract String getRedirectUri();
 
-  /**
-   * @return A key that is used to store authentication state in a session cookie.
-   */
+  /** @return A key that is used to store authentication state in a session cookie. */
   protected abstract String getSessionServiceKey();
 
   @Override
@@ -57,8 +55,7 @@ public abstract class OAuthCallbackServlet extends HttpServlet {
       return;
     }
 
-    var sessionOauthState =
-        (String) req.getSession().getAttribute(getSessionServiceKey());
+    var sessionOauthState = (String) req.getSession().getAttribute(getSessionServiceKey());
 
     if (!state.equals(sessionOauthState)) {
       res.setStatus(401);
--- a/server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
+++ b/server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
@@ -66,7 +66,7 @@ public class OAuthCallbackServlet extends HttpServlet {
     var tokenReqBody = new URLEncodedBuilder()
       .add("grant_type", "authorization_code")
       .add("code", code)
-      .add("redirct_uri", redirectUri)
+      .add("redirect_uri", redirectUri)
       .add("client_id", clientId)
       .add("client_secret", clientSecret);
 
--- a/server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
+++ b/server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
@@ -2,7 +2,6 @@ package com.google.musicanalysis.site;
 
 import java.io.IOException;
 import java.net.URI;
-import java.net.URLEncoder;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpRequest.BodyPublishers;
@@ -13,6 +12,7 @@ import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import com.google.gson.JsonParser;
+import com.google.musicanalysis.util.URLEncodedBuilder;
 
 @WebServlet("/api/oauth/callback")
 public class OAuthCallbackServlet extends HttpServlet {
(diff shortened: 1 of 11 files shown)


server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -37,40 +37,51 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
-    tokenRequestBody.append("&client_id=");
-    tokenRequestBody.append(URLEncoder.encode(Constants.SPOTIFY_CLIENT_ID, "UTF-8"));
-    tokenRequestBody.append("&client_secret=");
-    tokenRequestBody.append(URLEncoder.encode(spotifyClientSecret, "UTF-8"));
+    var tokenReqBody = new URLEncodedBuilder()
+      .add("grant_type", "authorization_code")
+      .add("code", code)
+      .add("redirct_uri", redirectUri)
+      .add("client_id", clientId)
+      .add("client_secret", clientSecret);
```
gracetian6, line 71: Oh is URLEncodedBuilder() a better function than StringBuilder()? (I see below that this is a custom class you have created to make things more convenient.)

server/src/main/java/com/google/musicanalysis/site/OAuthLoginServlet.java
```diff
@@ -2,40 +2,62 @@
...
+      res.setStatus(404);
+      return;
+    }
 
     // URI that user should be redirected to after logging in
     var redirectUri = System.getenv().get("OAUTH_CALLBACK_URI");
+    String oauthLoginUriBase;
+    URLEncodedBuilder oauthLoginUriParams;
+
+      // generate a nonce to prevent CSRF
```
seunomonije, line 31: nit: move comment back 
```suggestion
    // generate a nonce to prevent CSRF
```

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
laptou: Not better, but like you said, more convenient 😛

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -37,40 +37,51 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
     }
 
     var sessionOauthState = req.getSession().getAttribute("oauth:state");
+    var sessionOauthService = req.getSession().getAttribute("oauth:service");
```
geekster777, line 40: nit: you should only use `var` when the variable's type is explicitly declared when being initialized.

Ex:
```
var grid = new ArrayList<ArrayList<String>>();
```

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -37,40 +37,51 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
     if (!state.equals(sessionOauthState)) {
       res.setStatus(401);
       return;
     }
 
+    String clientSecret;
+    String clientId;
+    URI tokenUri;
+
+    if (sessionOauthService.equals("youtube")) {
```
geekster777, line 51: Having both the youtube and spotify oauth callbacks together in the same class hinders long-term maintenance since it becomes hard to detangle these two code paths. If you want to add a new spotify feature, you'll have to explicitly hide it from the youtube code path. Consider creating an abstract oauth class and having two separate routes/classes:
```
@WebServlet("/api/oauth/callback-spotify")
public class SpotifyOauthCallbackServlet extends OauthCallbackServlet {
  protected override clientSecret = System.getenv().get("SPOTIFY_CLIENT_SECRET");
  protected override clientId = SPOTIFY_CLIENT_ID;
  protected override tokenUri = URI.create("https://accounts.spotify.com/api/token");
  ...
}
```

This will give you more flexibility to branch more functionality as needed.

server/src/main/java/com/google/musicanalysis/site/OAuthLoginServlet.java
```diff
@@ -2,40 +2,62 @@
...
+    String oauthLoginUriBase;
+    URLEncodedBuilder oauthLoginUriParams;
+
+    // generate a nonce to prevent CSRF
+    var state = new BigInteger(130, new SecureRandom()).toString(32);
+    var session = req.getSession(true);
+    session.setAttribute("oauth:state", state);
+
+    if (path.equalsIgnoreCase("spotify")) {
+      session.setAttribute("oauth:service", "spotify");
```
geekster777, line 37: nit: use dashes as a separator for attributes instead of colons: `oauth-service`.

server/src/main/java/com/google/musicanalysis/util/URLEncodedBuilder.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.net.URLEncoder;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+
+/**
+ * A builder for creating URL-encoded query strings.
+ */
+public class URLEncodedBuilder {
```
geekster777, line 11: Nice util class! Surprised this didn't exist already.

server/src/main/java/com/google/musicanalysis/site/OAuthLoginServlet.java
```diff
@@ -2,40 +2,62 @@
...
     var redirectUri = System.getenv().get("OAUTH_CALLBACK_URI");
+    String oauthLoginUriBase;
+    URLEncodedBuilder oauthLoginUriParams;
+
+    // generate a nonce to prevent CSRF
+    var state = new BigInteger(130, new SecureRandom()).toString(32);
+    var session = req.getSession(true);
+    session.setAttribute("oauth:state", state);
+
+    if (path.equalsIgnoreCase("spotify")) {
```
geekster777, line 36: Same. This will also let you more explicitly declare your routes and leave a good abstraction in case you need any more oauth paths.

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -1,21 +1,37 @@
...
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import com.google.gson.JsonParser;
 
-@WebServlet("/api/oauth/callback")
-public class OAuthCallbackServlet extends HttpServlet {
+public abstract class OAuthCallbackServlet extends HttpServlet {
+  /** @return The name of this OAuth service. Used for storing session cookies and the like. */
+  public abstract String getServiceName();
```
geekster777, line 18: nit: make these protected.

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -36,41 +52,36 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
       return;
     }
 
-    var sessionOauthState = req.getSession().getAttribute("oauth:state");
+    var sessionOauthState =
+        (String) req.getSession().getAttribute("oauth-state-" + getServiceName());
```
geekster777, line 56: nit: remove some of the magic here and have a getStateSessionAttribute(). This will let you share some const `SPOTIFY_SESSION_ATTRUBUTE = "oauth-state-spotify"` between this getter and its setter.

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
geekster777: Same with other classes.

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
geekster777: It also creates a hard link so that updating one forces you to update the other.

PR #13, opened 2020-06-26
Change environment variables

laptou:
- It turns out that using `OAUTH_CALLBACK_URI` as an environment variable was prone to mistakes; I kept setting `OAUTH_REDIRECT_URI` instead and then scratching my head when I got an NPE
- There will probably be other things in the app that depend on the domain where the app is running, so now a `DOMAIN` env variable is used instead
- Secrets are stored in Google Secret Manager now instead of environment variables
  - This means that you have to be signed into gcloud in order for this to work correctly. Instructions are in server README


server/src/main/java/com/google/musicanalysis/site/OAuthLoginServlet.java
```diff
@@ -33,29 +52,25 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
     }
 
-    oauthLoginUriParams = oauthLoginUriParams
-        .add("response_type", "code")
-        .add("state", state)
-        .add("redirect_uri", redirectUri);
+    oauthLoginUriParams = oauthLoginUriParams.add("response_type", "code").add("state", state)
+        .add("redirect_uri", redirectUri.toString());
 
     // redirect to OAuth landing page
```
seunomonije, line 75: +1 Helpful comment

server/src/main/java/com/google/musicanalysis/site/OAuthLoginServlet.java
```diff
@@ -12,19 +15,35 @@
...
+      LOGGER.severe("The DOMAIN environment variable is invalid.");
+      res.setStatus(500);
+      return;
+    } catch (NullPointerException e) {
+      LOGGER.severe("The DOMAIN environment variable is not set.");
+      res.setStatus(500);
+      return;
+    }
+
+    var redirectUri = domainUri.resolve("/api/oauth/callback");
```
seunomonije, line 45: Used this link to look up `URI.resolve()` -> https://stackoverflow.com/questions/2534124/java-uri-resolve

Why is this necessary in this case? Asking for knowledge.

laptou: I always like to use methods that have special handling when I'm concatenating paths. In this case, I could just do a string concatenation and probably get away with it: "localhost:8080" + "/api/oauth/callback" = "localhost:8080/api/oauth/callback". But what if for some reason the first part is specified as "localhost:8080/"? That's technically the same path, but doing a string concatenation would give "localhost:8080//api/oauth/callback", which is not a valid path. `URI.resolve()` will handle this case automatically.

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -49,24 +54,39 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
     URI tokenUri;
 
     if (sessionOauthService.equals("youtube")) {
-      clientSecret = System.getenv().get("YOUTUBE_CLIENT_SECRET");
+      clientSecret = Secrets.getSecretString("YOUTUBE_CLIENT_SECRET");
```
gracetian6, line 57: yay

gracetian6, line 57: yay no more command typing

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -49,24 +54,39 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
+       LOGGER.severe("The DOMAIN environment variable is invalid.");
+       res.setStatus(500);
+       return;
+     } catch (NullPointerException e) {
+       LOGGER.severe("The DOMAIN environment variable is not set.");
+       res.setStatus(500);
+       return;
+     }
+
+    var redirectUri = domainUri.resolve("/api/oauth/callback");
```
gracetian6, line 84: THANK YOU FOR THESE ERROR MESSAGES!!

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -49,24 +54,39 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
 
     var tokenReqBody = new URLEncodedBuilder()
       .add("grant_type", "authorization_code")
       .add("code", code)
-      .add("redirct_uri", redirectUri)
+      .add("redirect_uri", redirectUri.toString())
```
gracetian6, line 89: wish I could've caught this typo :'(

gracetian6:
Thanks for README instructions! I'm able to get access tokens for both spotify and youtube 👍

server/.gcloudignore
```diff
@@ -0,0 +1,17 @@
...
+#
+.gcloudignore
+# If you would like to upload your .git directory, .gitignore file or files
+# from your .gitignore file, remove the corresponding line
+# below:
+.git
+.gitignore
+
+# Target directory for maven builds
+target/
```
geekster777, line 17: nit: add new line

server/src/main/java/com/google/musicanalysis/util/Secrets.java
```diff
@@ -0,0 +1,36 @@
...
+import java.nio.charset.StandardCharsets;
+import com.google.cloud.secretmanager.v1.SecretManagerServiceClient;
+import com.google.cloud.secretmanager.v1.SecretVersionName;
+
+public final class Secrets {
+  private static Object SYNC_OBJECT = new Object();
+  private static SecretManagerServiceClient CLIENT;
+
+  private static SecretManagerServiceClient getClient() throws IOException {
+    var localRef = CLIENT;
```
geekster777, line 13: nit: this localRef variable is unnecessary. You can just use CLIENT directly.

server/src/main/java/com/google/musicanalysis/util/Secrets.java
```diff
@@ -0,0 +1,36 @@
+package com.google.musicanalysis.util;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import com.google.cloud.secretmanager.v1.SecretManagerServiceClient;
+import com.google.cloud.secretmanager.v1.SecretVersionName;
+
+public final class Secrets {
+  private static Object SYNC_OBJECT = new Object();
```
geekster777, line 9: nit: make this `final`

server/src/main/java/com/google/musicanalysis/util/Secrets.java
```diff
@@ -0,0 +1,36 @@
...
+  private static SecretManagerServiceClient CLIENT;
```
geekster777, line 10: This isn't technically a constant (it's a singleton). It should use lowerCamelCase.

server/README.md
```diff
@@ -1,3 +1,36 @@
...
+4. Run `run.sh`
+
+## pitfalls
+
+- `run.sh` will assume that your server is running at `localhost:8080`. if you are running on Cloud
+  Shell, then you cannot access the website at `localhost:8080`. you can get around this by setting
+  the `DOMAIN` environment variable to the address of your Cloud Shell web preview instance, but
+  keep in mind that this address will change frequently.
+
+## faq
```
geekster777, line 28: Can you include instructions on how to add a Secret?

server/src/main/java/com/google/musicanalysis/util/Secrets.java
laptou: Isn't it possible for the value of `CLIENT` to change between the time we check whether it is null and the time we enter the `synchronized` block?

geekster777: Sure, but you're already checking if it's null a second time on lines 17 and 18

server/src/main/java/com/google/musicanalysis/site/YouTubeCallbackServlet.java
```diff
@@ -17,7 +24,12 @@ protected String getClientId() {
 
   @Override
   protected String getClientSecret() {
-    return System.getenv().get("YOUTUBE_CLIENT_SECRET");
+    try {
+      return Secrets.getSecretString("OUTUBE_CLIENT_SECRET");
```
gracetian6, line 28: bug -- please change to "YOUTUBE_CLIENT_SECRET" instead of "OUTUBE_CLIENT_SECRET"

server/src/main/java/com/google/musicanalysis/site/YouTubeCallbackServlet.java
```diff
@@ -27,7 +39,14 @@ protected String getTokenUri() {
 
   @Override
   protected String getRedirectUri() {
-    return System.getenv().get("YOUTUBE_CALLBACK_URI");
+    // URI that user should be redirected to after logging in
+    try {
+      var domainUri = URI.create(System.getenv().get("DOMAIN"));
+      return domainUri.resolve("/api/oauth/callback/youtube").toString();
```
gracetian6, line 45: Also I had some redirect_uri error when trying to do the oauth, so I had to add `/api/oauth/callback/youtube` and `api/oauth/callback/spotify` as possible redirect uris to the cloud project instead of only `api/oauth/callback`

laptou: yeah, I tried to add it myself but I couldn't find where it was

gracetian6: okay I changed it but didn't know if that was needed

PR #14, opened 2020-06-26
Musixmatch frontend

seunomonije:
Untested front-end code from yesterday. Thought y'all should take a look.


client/js/analysis.js
```diff
@@ -0,0 +1,26 @@
+
+window.onload = function() {
+    // Listen for on submit click
+    let formSubmit = document.getElementById("formSubmit");
+    formSubmit.addEventListener("click", fetchAndShowResponse, false);r
```
laptou, line 5: What's this `r`?

client/js/analysis.js
```diff
@@ -0,0 +1,26 @@
+
```
laptou, line 1: This file should be indented with 2 spaces

client/analysis.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
```
laptou, line 1: This file should be indented with 2 spaces

client/analysis.html
```diff
@@ -0,0 +1,26 @@
+
+
+<!DOCTYPE html>
+
+<html lang="en">
+    <head>
+        <meta charset="utf-8">
+        <meta name="description" content="STEP Capstone">
+        <script src="js/analysis.js"></script>
```
laptou, line 9: Consider moving the `script` tag to the end of your `body`. This way, your tag will have access to the DOM when it loads, instead of having to listen for a "page loaded" event.

seunomonije: will do.

jessicaslaughter:
You should get Grace's approval before requesting from me and Nick, just so the three of y'all get a lot of chances to review code before seeing our input. Feel free to ping anyone if you haven't gotten a review in 24 hours!

seunomonije:
I'm gonna try and merge this anyways because we can use it with Youtube instead

client/analysis.html
jessicaslaughter: Looks like this still needs to be resolved?

client/js/analysis.js
```diff
@@ -0,0 +1,27 @@
...
+    changeInnerText("response", value);
+    return value;
+}
+
+/**
+ * Change the contents of div
+ * @param elId the id of the element 
+ * @param value the value to insert as innerText
+ */
+function changeInnerText(elId, value) {
```
jessicaslaughter, line 24: Consider renaming this "setInnerText", and updating the function comment to be more generic since the element isn't necessarily a div (this helper could later be used for something other than the "response" div).

client/analysis.html
```diff
@@ -0,0 +1,26 @@
+
+
```
geekster777, line 2: Remove unnecessary whitespace

client/analysis.html
```diff
@@ -0,0 +1,26 @@
...
+        <meta name="description" content="STEP Capstone">
+        <script src="js/analysis.js"></script>
+        <link rel="stylesheet" href="css/analysis.css">
+        <title>Song Analysis</title>
+    </head>
+
+    <body>
+        <h1>Enter a song name and artist name</h1>
+
+        <p>Insert Song name</p>
```
geekster777, line 17: You should use `<label>`'s for the text describing an input.

client/js/analysis.js
geekster777: This was never completed

client/js/analysis.js
geekster777: +1. I'd also consider getting rid of this helper entirely since it makes the parent function less easy to read. In general, you probs don't need a helper function if it's under 3 lines of code (unless they're complicated lines of code).

client/js/analysis.js
geekster777: Please actually complete this. This file doesn't use 2-space indents and it should

client/analysis.html
geekster777: Please use 2 space indents

client/js/analysis.js
seunomonije: Ah my fault...I mixed up 2-space indents and 2 lines at the top of the file. My bad.

geekster777: Please remove the 2 lines at the top of the file as well :)

PR #15, opened 2020-06-29
Set up deployment

laptou:
- Fix and create config files so that website can be deployed successfully
  - Important note: the `server/src/main/webapp` directory is symlinked to the `client` directory


client/index.html
```diff
@@ -0,0 +1 @@
+oh lord jetson made another one
```
seunomonije, line 1: pack in the mail, it's gone

PR #16, opened 2020-06-29
Basic GDPR page and theming

laptou:
- Establish `base.css`: a stylesheet that should be included on every page and gives the website a theme
- Create basic GDPR analysis page with some information

Screenshot 2020-07-06 at 13 50 45 - Display 1


client/spotify-gdpr.html
```diff
@@ -0,0 +1,52 @@
...
+    <base target="_blank">
+    <link rel="stylesheet" href="css/base.css">
+    <link rel="stylesheet" href="css/spotify-gdpr.css">
+  </head>
+  <body>
+    <main>
+      <h1 id="title">
+        <span class="highlight-primary-1">Spotify GDPR Analysis</span>
+      </h1>
+      <span class="subtitle">See how you've changed</span>
```
gracetian6, line 16: if its not too inconvenient, could you add a screenshot in PR description? No need to do so now cause you showed it to us before 👍

PR #18, opened 2020-06-30
Displays topic details on liked videos via Youtube API

gracetian6:
- added sessions to store access token in `OAuthCallbackServlet.java`. Should be able to retrieve access token by calling parameter `spotify_access_token` or `youtube_access_token` in session. 
- makes HTTP request in `YoutubeServlet.java` to find liked videos of a user, and returns topic details (see pic below) 

**JSON output on `/api/youtube`**
image


server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse res) 
+        throws ServletException, IOException {
+        // don't know how to catch null pointer exception
+
+        // retrieve Youtube API key and access token 
+        String API_KEY = Secrets.getSecretString("YOUTUBE_API_KEY");
+        HttpSession session = req.getSession();
+        String accessToken = session.getAttribute("youtube_access_token").toString();
```
gracetian6, line 38: NEED HELP: if you don't log in to google oauth through `/api/oauth/login/youtube` then this page `/api/youtube` returns NullPointerException. Is there a way to detect this error in `/api/youtube` and then redirect immediately to login at `/api/oauth/login/youtube`? I tried doing try / catch on lines 35-36 but got tons of errors.

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -0,0 +1,59 @@
...
+            .add("myRating", "like")
+            .add("key", API_KEY);
+        URI youtubeUri = URI.create("https://www.googleapis.com/youtube/v3/videos?" + youtubeParam.build());
+
+        var httpClient = HttpClient.newHttpClient();
+        var youtubeReq = HttpRequest.newBuilder(youtubeUri)
+                .header("Authorization", "Bearer " + accessToken)
+                .header("Accept", "application/json")
+                .GET()
+                .build();
```
gracetian6, line 52: I'm making this HTTP request:
```
GET https://www.googleapis.com/youtube/v3/videos?part=topicDetails&myRating=like&key=[YOUR_API_KEY] HTTP/1.1

Authorization: Bearer [YOUR_ACCESS_TOKEN]
Accept: application/json```

File: server/pom.xml [hunk 3]
```diff
@@ -38,7 +38,28 @@
...
+      <version>v3-rev20200618-1.30.9</version>
+    </dependency>
+    
+    <!-- Google OAuth API -->
+    <dependency>
+      <groupId>com.google.apis</groupId>
+      <artifactId>google-api-services-oauth2</artifactId>
+      <version>v2-rev20200213-1.30.9</version>
+    </dependency>
+
```
laptou, line 62: Why did you include these dependencies? It doesn't seem like you are using them (instead you're just making all of the HTTP requests directly).

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -85,5 +87,9 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
 
     res.setContentType("text/html");
     res.getWriter().printf("<h1>the access token for %s is %s</h1>", getServiceName(), accessToken);
+    
+    // store access token in a session as youtube_access_token or spotify_access_token
+    HttpSession session = req.getSession();
+    session.setAttribute(getServiceName() + "_access_token", accessToken.toString());
```
laptou, line 93: Nick will tell you to use constants instead of doing a concatenation here (he calls it "magic"). See `getSessionServiceKey()`, make a similar method like `getSessionTokenKey()` that just returns `"oauth-token-youtube"` or `"oauth-token-spotify"`.

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
laptou: - You have to check if `session` is null. If the user hasn't created any session information yet, then it will be.
- You have to check if the result of `getAttribute` is null. If the user hasn't logged into YouTube, then it will be.
- Instead of returning a redirect, just return a 401 status code.

server/pom.xml
gracetian6: oh right let me get rid of them XD I added them cause I was going to use the java Youtube API package, but I'm not doing that anymore I'm just doing http requests

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
gracetian6: @laptou 
- I was able to do a try catch that throws 401 error for null pointer exception. I tried doing if statements, it was not enough and still triggered NullPointerException error
- about your last point on access token in a request header, does that mean adding it as `/api/youtube?accessToken=[access token here]`?

server/src/main/java/com/google/musicanalysis/util/Constants.java
```diff
@@ -6,4 +6,6 @@
       "343396465743-a46pc8ds6cquiatmdnvf88o6aaf3jcdo.apps.googleusercontent.com";
   public static final String SPOTIFY_SESSION_KEY = "oauth-service-spotify";
   public static final String YOUTUBE_SESSION_KEY = "oauth-service-youtube";
+  public static final String YOUTUBE_SESSION_TOKEN_KEY = "youtube_access_token";
+  public static final String SPOTIFY_SESSION_TOKEN_KEY = "spotify_access_token";
```
laptou, line 10: these should all have a consistent format; make them all `kebab-case` instead of `snake_case` and prefix them with `oauth-`

server/pom.xml
```diff
@@ -38,7 +38,7 @@
         <artifactId>gson</artifactId>
         <version>2.8.6</version>
     </dependency>
-
+
```
jessicaslaughter, line 41: nit: remove extra line

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -0,0 +1,58 @@
+/**
+ * Sample Java code for youtube.videos.list
+ * See instructions for running these code samples locally:
+ * https://developers.google.com/explorer-help/guides/code_samples#java
+ */
```
jessicaslaughter, line 5: nit: remove this comment

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -35,6 +38,11 @@
   /** @return A key that is used to store authentication state in a session cookie. */
   protected abstract String getSessionServiceKey();
 
+  /** @return A key that is used to store authentication state in a session cookie. */
+  protected String getSessionTokenKey() {
+    return Constants.YOUTUBE_SESSION_TOKEN_KEY;
+  }
```
geekster777, line 44: This should exist in the child class, not here. There should be no references to Youtube in this file.

geekster777: This function should either be abstract or only exist in the Youtube file.

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
```diff
@@ -85,5 +93,9 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
 
     res.setContentType("text/html");
     res.getWriter().printf("<h1>the access token for %s is %s</h1>", getServiceName(), accessToken);
+    
+    // store access token in a session as youtube_access_token or spotify_access_token
+    HttpSession session = req.getSession();
+    session.setAttribute(getSessionTokenKey(), accessToken.toString());
```
geekster777, line 99: 2 things:
1, this should be `this.getSessionTokenKey()`

2, Is there a reference to `getSessionTokenKey()` in the Spotify OAuth Servlet? If not, this will likely introduce errors or unintended side effects. You should update that file as well.

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -0,0 +1,58 @@
...
+@WebServlet("/api/youtube")
+public class YoutubeServlet extends HttpServlet {
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse res) 
+        throws ServletException, IOException {
+        // retrieve Youtube API key and access token 
+        String API_KEY = Secrets.getSecretString("YOUTUBE_API_KEY");
+        HttpSession session = req.getSession();
+        var accessToken = session.getAttribute("oauth-access-token-youtube");
+        if (session == null || accessToken == null) {
```
geekster777, line 34: You don't need to check if `session` is null here. It would have already thrown an error on line 33 if it was.

server/pom.xml
gracetian6: image
Everything is spaced so that there's a line between each dependency, so removing an extra line would be imbalanced?

jessicaslaughter: You can leave it here then, but since the only change in that file was a new line, it's not really in scope for this PR (looks like it was leftover from changes that got reverted in this PR earlier).

server/src/main/java/com/google/musicanalysis/site/OAuthCallbackServlet.java
gracetian6: Fixed `getSessionTokenKey()`. `this.getSessionTokenKey()` and `getSessionTokenKey()` both work? Other functions on the top of this `OauthCallbackServlet.java` file are also called without calling `this.` such as `getRedirectUri()`, `getClientId()`, ... `getServiceName()` within this file.

gracetian6:
@laptou Help on fixing lint error? The lint error says there's an extra line in base.css, but there is not base.css file so no idea how to fix it.

When I tried merging the `fix-css-lint-1` branch where you fixed lint error, then the server didn't work in `api/oauth/login/youtube` etc. So I don't want to merge that branch.

gracetian6:
The lint error I have will be fixed once this PR https://github.com/googleinterns/step99-2020/pull/21 gets merged into master.

PR #19, opened 2020-06-30
Naturallang backend

seunomonije:
Very similar to perspective backend. Also fixed file directory issues.

image


server/src/main/java/com/google/musicanalysis/api/naturallanguage/SentimentJsonBuilder.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.google.gson.JsonObject;
+
+public class SentimentJsonBuilder {
+
+  private String text;
+
+  public SentimentJsonBuilder(String text) {
+    this.text = text;
+  }
```
geekster777, line 11: Will this class ever have more info in it? If not, just make this a function rather than a class. You don't need to store this text when I expect this class will always get used like:
```
return new SentimentJsonBuilder(text).buildJson();
```

geekster777: Consider making this a private method of `SentimentRequest`

PR #20, opened 2020-07-01
Contiguous backend

seunomonije:
Messed up the API key again. It's deleted from gcloud now so it's meaningless. 

### Contiguous backend
Successfully integrated all 3 APIs into one servlet. Can now get all information from a song with just a song name and an artist. 
Screenshot 2020-07-01 at 2 05 58 PM


server/src/main/java/com/google/musicanalysis/api/naturallanguage/SentimentRequest.java
```diff
@@ -22,7 +22,9 @@ public SentimentRequest(String text) {
    * @return the JSON response as string
    */
   public String getResponse() throws MalformedURLException, IOException {
-    URL url = new URL("https://language.googleapis.com/v1/documents:analyzeSentiment?key=");
+    URL url =
+        new URL(
+            "https://language.googleapis.com/v1/documents:analyzeSentiment?key=");
```
gracetian6, line 27: formatting looks a little weird - could you put 26-27 on the same line? (or does it exceed char limit)

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,157 @@
...
+
+    res.getWriter().write("<h3>NLP</h3>");
+    MagnitudeAndScore nlpObject = analyzeWithNLP(trimmedLyrics);
+    res.getWriter().println("Magnitude: " + nlpObject.magnitude + "Score: " + nlpObject.score);
+  }
+
+  private class MagnitudeAndScore {
+    Double magnitude;
+    Double score;
+
```
gracetian6, line 53: seun really churning out all this hundred line code XD

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,157 @@
...
+    public MagnitudeAndScore(Double magnitude, Double score) {
+      this.magnitude = magnitude;
+      this.score = score;
+    }
+  }
+
+  private MagnitudeAndScore analyzeWithNLP(String text) throws IOException {
```
gracetian6, line 60: could call this `analyzeNLP()`? but no big deal.

Also please add function comments in line 59

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,157 @@
...
+    MagnitudeAndScore nlpResults =
+        new MagnitudeAndScore(
+            Double.valueOf(sentimentObject.get("magnitude").toString()),
+            Double.valueOf(sentimentObject.get("score").toString()));
+
+    return nlpResults;
+  }
+
+  private HashMap<String, String> analyzeWithPerspective(String text) throws IOException {
+
```
gracetian6, line 79: same comment as above -- add fn comment and change to name `analyzePerspective()`

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,157 @@
...
+              .getAsJsonObject("summaryScore")
+              .get("value")
+              .toString());
+    }
+
+    return perspectiveResults;
+  }
+
+  private String getLyrics(String trackName, String artistName) throws IOException {
+
```
gracetian6, line 116: same fn comment as before

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,157 @@
...
+                "FLIRTATION"));
+
+    // Calling the API
+    PerspectiveRequest perspectiveRequest = new PerspectiveRequest(text, attributes);
+    String response = perspectiveRequest.getResponse();
+
+    // Extracting the results
+    JsonElement jElement = JsonParser.parseString(response);
+    JsonObject jObject = jElement.getAsJsonObject();
+    JsonObject attributeObject = jObject.getAsJsonObject("attributeScores");
```
gracetian6, line 98: I see this section that "extracts the results" is repeated  like in lines 65-68 above except for the input parameter `"attributeScores"`, is it possible to create a helper fn that reduces duplicate code?

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,157 @@
...
+    JsonElement jElement = JsonParser.parseString(response);
+    JsonObject jObject = jElement.getAsJsonObject();
+    JsonArray resultList =
+        jObject.getAsJsonObject("message").getAsJsonObject("body").getAsJsonArray("track_list");
+    Iterator<JsonElement> iterator = resultList.iterator();
+    JsonElement firstResult = iterator.next();
+    String trackId =
+        firstResult.getAsJsonObject().getAsJsonObject("track").get("track_id").toString();
+
+    /** Getting the lyrics from the extracted id */
```
gracetian6, line 134: you have some hefty `/**comments */` that do substantial parts of the code detailed in lines 115 and 134. Could you possibly create helper functions in each case? This could make the code more readable.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,157 @@
...
+
+    // Calling the API
+    params = new MusixParamBuilder("track.lyrics.get", trackId);
+    musixReq = new MusixRequest("track.lyrics.get", params.filterParamString());
+    response = musixReq.grabResponse();
```
gracetian6, line 139: Note that these lines that "call the API" are almost identical to lines 120-122.  Maybe add a helper function that calls the API?

So, this way in your helper functions that (1) gets track id from musixmatch and (2) get lyrics from extracted id, you just have to call this helper function that "calls the API" in both (1) and (2).

seunomonije:
Made some changes. Musix match stuff is now gone and youtube stuff is in. As of right now gets all the comments given a video name, taking the most popular search result as the match. Here is an example of it in action.

Screenshot 2020-07-07 at 1 18 43 PM
Screenshot 2020-07-07 at 1 09 50 PM

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
seunomonije: I see what you're saying here...the only reason why I'm hesitant is that sometimes I need to getAsJsonObject(), and sometimes I need to getAsJsonArray().

seunomonije: That would require me:
1. Having a JsonObject helper function and a Json Array helper
or 
2. Having some sort of extra value I pass in that determines wheher to get as Array or get as Object. 

I'm not opposed to it but I'm not sure that I would use the helper more than twice. However I am biased towards my own code lol.

gracetian6: okay, got it! Then my request doesn't seem to helpful. Thank you taking a step back to think about your code.

server/src/main/java/com/google/musicanalysis/api/youtube/YoutubeRequest.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+
+public class YoutubeRequest {
```
jessicaslaughter, line 13: This class should have a Javadoc comment

server/src/main/java/com/google/musicanalysis/api/youtube/YoutubeRequest.java
```diff
@@ -0,0 +1,77 @@
...
+  private ArrayList<String> parameters = new ArrayList<String>();
+
+  public YoutubeRequest(String operation, String... parameters) {
+    this.operation = operation;
+    for (String parameter : parameters) {
+      this.parameters.add(parameter);
+    }
+  }
+
+  public String getResult() throws MalformedURLException, IOException {
```
jessicaslaughter, line 27: needs a comment

server/src/main/java/com/google/musicanalysis/api/youtube/YoutubeRequest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    String urlString = BASE_URL + this.operation + parameterString + apiKey;
+    URL url = new URL(urlString);
+    return url;
+  }
+
+  private String buildParams(ArrayList<String> ops) {
+    // More cases can be added in the future to this
+    switch (this.parameters.size()) {
```
jessicaslaughter, line 64: Instead of adding a new cases, this would scale better as a loop that just adds onto an existing parameter string for each item in the ArrayList.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,155 @@
...
+      Map.Entry entry = (Map.Entry) mapIterator.next();
+      res.getWriter().println(entry.getKey() + ": " + entry.getValue());
+    }
+
+    res.getWriter().write("<h3>NLP</h3>");
+    MagnitudeAndScore nlpObject = analyzeWithNLP(comment);
+    res.getWriter().println("Magnitude: " + nlpObject.magnitude + "Score: " + nlpObject.score);
+  }
+
+  private class MagnitudeAndScore {
```
jessicaslaughter, line 55: Consider renaming this something like NLPResult. The current name isn't flexible for any future changes to this class.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * Analyzes the given string with the Natural Language API.
+   *
+   * @param text The text that will be analyzed by the Natural Language API.
+   * @return A MagnitudeAndScore object with the results
+   */
+  private MagnitudeAndScore analyzeWithNLP(String text) throws IOException {
+
+    // Calling the API
```
jessicaslaughter, line 73: This comment isn't really needed, the code is self-explanatory

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,155 @@
...
+    PerspectiveRequest perspectiveRequest = new PerspectiveRequest(text, attributes);
+    String response = perspectiveRequest.getResponse();
+
+    // Extracting the results
+    JsonElement jElement = JsonParser.parseString(response);
+    JsonObject jObject = jElement.getAsJsonObject();
+    JsonObject attributeObject = jObject.getAsJsonObject("attributeScores");
+
+    // Placing them in a Hash Map
```
jessicaslaughter, line 117: Comment isn't needed, code is self-explanatory

server/src/main/java/com/google/musicanalysis/api/youtube/YoutubeRequest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    in.close();
+
+    return buffer.toString();
+  }
+
+  private URL buildUrl() throws MalformedURLException {
+    ArrayList<String> searchOps = new ArrayList<String>(Arrays.asList("?q=", "&videoCaption="));
+    ArrayList<String> commentOps = new ArrayList<String>(Arrays.asList("?part=", "&videoId="));
```
geekster777, line 48: You should be using a map of parameters instead of an array. This is currently brittle since it relies on people remembering to pass in parameters in a specific order and for them to match up with these array lists. Instead, you can accept a map of parameters, and do something like this:
```
// javascript code. Sorry, I'm a java scrub.
function build(paramMap) {
  const parts = Object.keys(paramMap).map(key => {
    return `${encode(key)}=${encode(paramMap[key])}`;
  );
  return `${url}?${parts.join('&')}`;
}
```

geekster777: This will also future-proof your buildParams function.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,155 @@
...
+    ArrayList<String> searchResults = new ArrayList<>();
+
+    // Traversing to the items array
+    JsonElement jElement = JsonParser.parseString(response);
+    JsonObject jObject = jElement.getAsJsonObject();
+    JsonArray itemsArray = jObject.getAsJsonArray("items");
+
+    for (JsonElement el : itemsArray) {
+
+      // Traversing each item in the array
```
geekster777, line 142: This comment is unnecessary. Better to say what you're doing with the items

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,155 @@
...
+              .toString());
+    }
+
+    return perspectiveResults;
+  }
+
+  private String getVideoId(String response) {
+    ArrayList<String> searchResults = new ArrayList<>();
+
+    // Traversing to the items array
```
geekster777, line 135: this is false.

server/src/main/java/com/google/musicanalysis/api/youtube/YoutubeRequest.java
seunomonije: This is dope. Don't even need the buildParams function anymore.

server/src/main/java/com/google/musicanalysis/api/youtube/YoutubeRequest.java
seunomonije: Ended up getting rid of this fn

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -0,0 +1,159 @@
...
+    String videoName = "seun omonije";
+
+    // Use like this: {url_parameter, value}
+    HashMap<String, String> args = new HashMap<>();
+
+    args.put("q", videoName);
+    String videoIdJson = new YoutubeRequest("search", args).getResult();
+    String videoId = getVideoId(videoIdJson);
+
+    args.remove("q");
```
geekster777, line 37: Make a new variable instead of re-using a generic one. You can then name the previous one videoArgs and this one commentsArgs. This will make your code more readable and easier to grow.

PR #22, opened 2020-07-07
Add chart of song positions

laptou:
- Add chart of song positions
- Add styles to polish the chart
- Add varied dummy data to test the chart properly


client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,198 @@
...
+    case 'replace': {
+      const {oldItem, newItem} = change;
+
+      const idx = indices.get(newItem);
+      current[idx] = oldItem;
+
+      indices.set(newItem, null);
+      indices.set(oldItem, idx);
+      break;
+    }
```
gracetian6, line 116: could you add some comments? not really sure what's going on in chunks of code like this.

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,198 @@
...
+      }
+
+      indexHistory.push(index);
+      indicesHistory.set(id, indexHistory);
+    }
+  }
+
+  x.push(new Date(+x[x.length - 1] - 86400000));
+  generation++;
+}
```
gracetian6, line 140: a comment for this block would be nice too

client/js/spotify-gdpr.js
laptou: you're right, i apologise

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
...
+    case 'move': {
+      const {from, to} = change;
+
+      const a = current[from];
+      const b = current[to];
+      current[from] = b;
+      current[to] = a;
+
+      indices.set(a, to);
+      indices.set(b, from);
```
gracetian6, line 105: okay nice you are swapping indices, and storing the new change

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
...
+  x.push(new Date(+x[x.length - 1] - 86400000));
+  generation++;
+}
+
+// format of a chart series in c3 is ['name of series', point1, point2, ...]
+const data =
+  /** @type {Array<[string, ...number[]]>} */
+  (Array
+    .from(indicesHistory.entries())
+    .map(([id, history]) => [id, ...history]));
```
gracetian6, line 151: great comment!!

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
...
+        indexHistory.push(null);
+      }
+
+      indexHistory.push(index);
+      indicesHistory.set(id, indexHistory);
+    }
+  }
+
+  // x-value of this on the chart is the current date minus 86,400,000ms (1 day)
+  x.push(new Date(+x[x.length - 1] - 86400000));
```
gracetian6, line 142: set a constant equal to `86400000`? nick doesn't like magic numbers

laptou: i don't think the number of milliseconds in a day is a "magic number", that's like saying 24 hours in a day should be abstracted into a constant

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
...
+// indices)
+const indicesHistory = new Map(current.map((id, idx) => [id, [idx]]));
+
+// for each changeset (set of changes that occurred on a day) apply the opposite
+// change so that we can reconstruct the ordering of the top songs list on that
+// day
+for (const changeSet of changes.reverse()) {
+  for (const change of changeSet.reverse()) {
+    switch (change.type) {
+    case 'move': {
```
seunomonije, line 96: nit: indent

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
+// @ts-check
+/* global c3 */
+
+// most recent ordering of top songs
+const latest = [
+  'song3',
```
geekster777, line 6: Why are songs 4, 6, 8, and 9 missing from this list. Also can you add a clarifying comment why only songs 0-9 are out of order and the rest are linear?

geekster777: Follow up, it's looking like this file is a bunch of static mock data. If that's the case, it would be helpful to have a file level comment explaining this.

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
...
+  [{type: 'move', from: 20, to: 21}],
+  [{type: 'move', from: 12, to: 13}],
+  [],
+  [],
+  [],
+];
+
+// number of days we have backtracked
+let generation = 1;
+const x = [new Date()];
```
geekster777, line 80: What is x? Please use a more descriptive name

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
...
+// current position of each song (map from song id to current index)
+const indices = new Map(current.map((id, idx) => [id, idx]));
+// historical position of each song (map from song id to array of historical
+// indices)
+const indicesHistory = new Map(current.map((id, idx) => [id, [idx]]));
+
+// for each changeset (set of changes that occurred on a day) apply the opposite
+// change so that we can reconstruct the ordering of the top songs list on that
+// day
+for (const changeSet of changes.reverse()) {
```
geekster777, line 93: `changes.reverse()` reverses the contents of the array in place. I suspect the two for loops here can be consolidated down into one:
```
const changeHistory = changes.slice().reverse();
for (const change of changeHistory) {
   ...
}
```

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
...
+  for (const change of changeSet.reverse()) {
+    switch (change.type) {
+      case 'move': {
+        const {from, to} = change;
+
+        const a = current[from];
+        const b = current[to];
```
geekster777, line 100: nit: you should use more descriptive variable names here. Consider using `fromSong` and `toSong`.

client/js/spotify-gdpr.js
geekster777: This should be a constant, and written as an equation.
```
const MILLISECONDS_PER_DAY = 24 * 60 * 60 * 1000
```

It's pretty common to see HOURS_PER_DAY as a constant.

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
...
+      for (let i = 0; i < generation; i++) {
+        indexHistory.push(null);
+      }
+
+      indexHistory.push(index);
+      indicesHistory.set(id, indexHistory);
+    }
+  }
+
+  // x-value of this on the chart is the current date minus 86,400,000ms (1 day)
```
geekster777, line 141: I see the name `x` is corresponding to the x-axis values in your graph. Unfortunately `x` is colloquially assumed to be a generic name. Consider using something like `deltas` or `xAxisValues`

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
...
+  x.push(new Date(+x[x.length - 1] - 86400000));
+  generation++;
+}
+
+// format of a chart series in c3 is ['name of series', point1, point2, ...]
+const data =
```
geekster777, line 147: nit: call this `chartData`. `data` is a bit too generic.

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,203 @@
...
+      .from(indicesHistory.entries())
+      .map(([id, history]) => [id, ...history]));
+
+const dateFormat = new Intl.DateTimeFormat('en', {
+  day: 'numeric',
+  month: 'short',
+  year: 'numeric',
+});
+
+const chart = c3.generate({
```
geekster777, line 159: This file should be lightly refactored into functions. This should be pretty easy since all your code is currently logically and sequentially grouped. But I imagine it looking something like:
```
function getGdprData() {
  // currently mock data
}

function getChanges() {
  // currently mock data;
}

function applyChanges(gdprData, changes, histogram) {
}

// ... etc.
```

This should help you as you further flesh out this file. I suspect you may have already done something similar in a future PR, in which case this code is easy enough to follow that we can defer until then.

client/js/spotify-gdpr.js
laptou: That won't work because `changes` is an array of arrays. And it doesn't matter that the array is reversed in-place because I'm only iterating over it once, and the reversed array is returned.

geekster777: Ah, my mistake. Then you should either modify your original arrays on a separate line, or clone your arrays first in your loops:
```
changes.reverse();
for (changeSet of changes) {
  changeSet.reverse();
  for (change of changeSet) {
    ...
  }
}

// or
for (changeSet of changes.slice().reverse()) {
  for (change of changeSet.slice().reverse()) {
    ...
  }
}
}
```

PR #23, opened 2020-07-07
Add GDPR record unzipping

laptou:
- Include [zip.js](https://gildas-lormeau.github.io/zip.js/) library for unzipping JS files
  - This accounts for a large portion of the 5,000 lines in this PR; there doesn't appear to be a CDN that hosts these JS files so I just copied them into the repo
- Create file upload for uploading zip files
- Read GDPR records from said files
- Dump them to the `console`

### DIFF
--- a/client/js/zip/z-worker.js
+++ b/client/js/zip/z-worker.js
@@ -1,0 +1,153 @@
+/* jshint worker:true */
+(function main(global) {
+	"use strict";
+
+	if (global.zWorkerInitialized)
+		throw new Error('z-worker.js should be run only once');
+	global.zWorkerInitialized = true;
+
+	addEventListener("message", function(event) {
+		var message = event.data, type = message.type, sn = message.sn;
+		var handler = handlers[type];
+		if (handler) {
+			try {
+				handler(message);
+			} catch (e) {
+				onError(type, sn, e);
+			}
+		}
+		//for debug
+		//postMessage({type: 'echo', originalType: type, sn: sn});
+	});
+
+	var handlers = {
+		importScripts: doImportScripts,
+		newTask: newTask,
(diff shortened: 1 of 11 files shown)


gracetian6:
HELP this PR has 200+ lines of code to review 😨

client/js/zip/z-worker.js
```diff
@@ -0,0 +1,153 @@
+/* jshint worker:true */
+(function main(global) {
+	"use strict";
+
+	if (global.zWorkerInitialized)
+		throw new Error('z-worker.js should be run only once');
+	global.zWorkerInitialized = true;
```
gracetian6, line 7: This is 150+ lines with vague/no comments. please add comments to `z-worker.js` bc I'm illiterate without comments

laptou: This file is part of zip.js. I didn't even write this; like I said, it's a library that I just copied into the repo because there is no CDN to serve it. You do not have to review any files in the `zip/` folder, because I did not write them and will not be editing them either.

gracetian6: Okay I see, I was confused.

client/js/zip/z-worker.js
```diff
@@ -0,0 +1,153 @@
+/* jshint worker:true */
+(function main(global) {
```
gracetian6, line 2: add a quick comment at the top saying `/* taken from / part of zip.js library ` so its clear? You have a good block comment in the `zip.js` file, so doing the same here would be great.

jessicaslaughter:
Ideally to make this PR less confusing, the zip.js files should've been added to the repo in a single PR, and the work using zip.js would go in another. Just a note for future PR's.

PR #24, opened 2020-07-07
Add collation of GDPR records

laptou:
- Adds a function which combines & collates the records extracted by `getStreamingData` into a `Map` that can easily be used as input to the chart


client/js/analysis/gdpr.js
```diff
@@ -11,9 +11,63 @@
...
+ * @template V
+ * @param {Map<K, V>} map The map to update.
+ * @param {K} key The key associated with the value to update.
+ * @param {(value: V | undefined, key: K) => V} fn Returns the new value to be
+ * associated with `key` in `map`.
+ * @returns {Map<K, V>} `map`
+ */
+function update(map, key, fn) {
+  const value = map.get(key);
+  map.set(key, fn(value, key));
```
seunomonije, line 66: I'm assuming the `fn(value, key)` you put here handles the key in map case as well as key is not in map. If that's the case, then why do you need the if statement in the `getOrDefault()` function?

client/js/analysis/gdpr.js
```diff
@@ -67,14 +121,95 @@ export async function getStreamingData(data) {
...
+  /** @type {ArtistTrackTimeMap} */
+  let window = new Map();
+
+  while (index < data.length) {
+    while (currentRecord && +currentRecord.endTime < windowStart + windowSize) {
+      const {artistName, trackName, msPlayed} = currentRecord;
+
+      const windowArtistTracks = getOrDefault(
+        window, artistName, () => new Map(),
+      );
```
seunomonije, line 189: So basically if there's key is undefined you make a new map. I don't know if you plan to have different functions inside, and it would make sense if you plan to use it later. Otherwise I suggest just making the new map in the actual function from a reader's point of view.

client/js/analysis/gdpr.js
```diff
@@ -67,14 +121,95 @@ export async function getStreamingData(data) {
...
+      update(
+        windowArtistTracks,
+        trackName,
+        (windowTrackTime = 0) => windowTrackTime + msPlayed,
+      );
+
+      update(
+        totalArtistTracks,
+        trackName,
+        (totalTrackTime = 0) => totalTrackTime + msPlayed,
```
seunomonije, line 203: Same comments as before. I see what you're doing tho.

client/js/analysis/gdpr.js
laptou: makes sense; i wrote these functions and then deleted them and then put them back b/c i wasn't sure whether they were actually necessary or not. i'll just get rid of them

client/js/analysis/gdpr.js
gracetian6: I second ^^

gracetian6:
hmm personally I liked the helper functions cause there was less lines, but both make sense.

client/js/analysis/gdpr.js
```diff
@@ -11,9 +11,63 @@
...
+ * @property {Date} endTime Date and time of when the stream ended in UTC
+ * format.
+ * @property {string} artistName Name of "creator" for each stream (e.g. the
+ * artist name if a music track).
+ * @property {string} trackName Name of items listened to or watched (e.g. title
+ * of music track or name of video).
+ * @property {number} msPlayed How many mili-seconds the
+ * track was listened.
+ */
+
```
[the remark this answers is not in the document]
laptou: It's a `typedef`. It describes a type of object, kind of like an interface. In this case, it defines what a GDPR record object should look like, what properties it has, what the types of those properties are, and what each property means. This allows you to know what I'm talking about when I write `/** @type {GDPRRecord} */` JS doesn't have strong typing by default, but I'll be damned if I have to code blindly without any types at all.

client/js/analysis/gdpr.js
```diff
@@ -67,14 +121,95 @@ export async function getStreamingData(data) {
...
       (error) => reject(error),
     );
   });
 }
+
+/** @typedef {Map<string, Map<string, number>>} ArtistTrackTimeMap */
+/** @typedef {{
+ * start: Date,
+ * totals: ArtistTrackTimeMap
+ * }} ArtistTrackTimeWindow */
```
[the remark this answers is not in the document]
gracetian6: ok I understand now

client/js/analysis/gdpr.js
```diff
@@ -67,14 +79,117 @@ export async function getStreamingData(data) {
...
+
+  // a map of maps: key is artist name, value is a map where key
+  // is track name and value is time listened to that song in total
+  /** @type {ArtistTrackTimeMap} */
+  const totals = new Map();
+
+  // each window is a map of maps: key is artist name, value is a map where key
+  // is track name and value is time listened to that song during this window
+  /** @type {ArtistTrackTimeMap} */
+  let window = new Map();
```
jessicaslaughter, line 143: "window" is a bit confusing of a name, because here it's an ArtistTrackTimeMap but "windows" is an array of ArtistTrackTimeWindow's (what I'm assuming to be an actual "window"). Consider renaming to something like "windowTotals" to make the while loop below more readable.

client/js/spotify-gdpr.js
```diff
@@ -1,5 +1,40 @@
...
+      ),
+  );
+
+  if (!inputUpload.files) return;
+
+  const fileUpload = inputUpload.files.item(0);
+  if (!fileUpload) return;
+
+  const records = await getStreamingData(fileUpload);
+  console.log(records);
```
jessicaslaughter, line 33: remove debug statements

client/js/spotify-gdpr.js
```diff
@@ -1,5 +1,40 @@
...
+
+  const collatedRecords = collateStreamingData(records);
+  console.log(collatedRecords);
```
jessicaslaughter, line 36: remove debug statements

client/jsconfig.json
```diff
@@ -3,7 +3,9 @@
     "checkJs": true,
     "target": "es2020",
     "module": "es2020",
-    "strict": true
+    "strictPropertyInitialization": true,
+    "strictNullChecks": true,
+    "strictBindCallApply": true
```
jessicaslaughter, line 8: Same here, these types of changes should go in a separate PR in the future, since they're not relevant to the purpose of this PR.

client/.eslintrc
```diff
@@ -58,7 +58,8 @@
           "message": "@extends is to be used over @augments as it is more evocative of classes than @augments",
           "replacement": "extends"
         }
-      }
+      },
+      "mode": "typescript"
```
jessicaslaughter, line 62: This type of change should go in a separate PR in the future since it's not relevant to this PR's purpose.

client/js/analysis/gdpr.js
```diff
@@ -67,14 +79,117 @@ export async function getStreamingData(data) {
...
 }
+
+/** @typedef {Map<string, Map<string, number>>} ArtistTrackTimeMap */
+/** @typedef {{
+ * start: Date,
+ * totals: ArtistTrackTimeMap
+ * }} ArtistTrackTimeWindow */
+
+/**
+ * Calculates aggregates from a list of GDPR records.
```
jessicaslaughter, line 114: Be more specific in this sentence, aggregates of what? The typedefs and return value give those details in a technical sense, but this comment should still give a reader enough detail to understand what this function does without having to piece together the values of the nested maps into a summary.

client/js/analysis/gdpr.js
```diff
@@ -67,14 +79,117 @@ export async function getStreamingData(data) {
...
                 .all(contentPromises)
                 .then((contentWithIndices) => {
-                  const content = contentWithIndices
-                      .map((entry) => entry.data)
-                      .reduce((dataA, dataB) => dataA.concat(dataB), []);
-                  resolve(content);
+                  resolve(
+                      contentWithIndices
+                          .map((entry) => entry.data)
+                          .reduce((dataA, dataB) => dataA.concat(dataB), [])
```
geekster777, line 85: nit: use `accumulator` and `data` instead of `a` and `b` when using reduce functions.

client/js/analysis/gdpr.js
```diff
@@ -67,14 +79,117 @@ export async function getStreamingData(data) {
...
+                          .map((entry) => {
+                            // dates are strings in JSON, convert to JS date
+                            const endTime = new Date(entry.endTime);
+
+                            // move according to UTC offset
+                            const endTimeMs = +endTime;
+                            const endTimeOffset =
```
geekster777, line 92: nit: name this endTimeOffsetMs to let folks know how you're intending to change the units of getTimezoneOffset. Unfortunately the fact that getTimezoneOffset returns a number in minutes isn't clear from the API name, but variable names like this help clarify this as well as your intentions.

client/js/analysis/gdpr.js
```diff
@@ -67,14 +79,117 @@ export async function getStreamingData(data) {
...
+ * keys are artists and whose values are maps from track names to duration
+ * listened. `windows` is an array of window objects. Each window object has a
+ * `start` representing the start of that window, and a `totals`, representing
+ * the durations listened to each song during that window.
+ */
+export function collateStreamingData(data) {
+  let index = 0;
+  let currentRecord = data[0];
+
+  const WINDOW_SIZE = 86400000;
```
geekster777, line 128: nit: make this `24 * 60 * 60 * 1000`

client/js/analysis/gdpr.js
```diff
@@ -67,14 +79,117 @@ export async function getStreamingData(data) {
...
+  // round down to the start of the day of the first record
```
geekster777, line 129: This is a good example of a comment that explains the intention of the code without explaining the mechanics of it. Nice!

client/js/analysis/gdpr.js
geekster777, line 143: consider using a different name than window to eliminate namespace collisions with the javascript window object. This is mostly for developer sanity, since most developers will assume `window.value` refers to the global window object instead of your local variable. Similarly if they ever try to modify your code, they may not notice this variable and will assume they have access to the global window object when they don't.

Consider using variable names like `timeframes`

laptou: I subbed out the term "window" for "interval".

client/js/spotify-gdpr.js
laptou: if i don't have these then the variable isn't used for anything (which will trip off eslint)
also this pull request doesn't even represent finished code, `client-style-5` will remove the console logs
is that okay?

client/js/analysis/gdpr.js
```diff
@@ -67,14 +79,120 @@ export async function getStreamingData(data) {
...
+
+      if (!totalTrackTime) {
+        totalTrackTime = 0;
+      }
+
+      totalTrackTime += msPlayed;
+
+      totalArtistTracks.set(trackName, totalTrackTime);
+
+      currentRecord = data[++index];
```
geekster777, line 189: nit: separate the array index access and the incrementation. These should be on separate lines to improve readability.

client/js/spotify-gdpr.js
geekster777: It's common to use `delete` in these cases:
```
// Records are currently unused.
delete records;
```

laptou: ESLint says: `Parsing error: Deleting local variable in strict mode`, and my IDE highlights everything in red when I try to do that...

laptou: Instead, I just put an ESLint ignore comment

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,271 @@
...
+    [],
+    [{type: 'move', from: 20, to: 21}],
+    [{type: 'move', from: 12, to: 13}],
+    [],
+    [],
+    [],
+  ];
+}
+
+const MS_PER_DAY = 24 * 60 * 60 * 100;
```
geekster777, line 125: nit: this should be `* 1000`

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,271 @@
...
+ * Populates `chart` with historical track data derived from `latest` and
+ * `changes`.
+ *
+ * @param {string[]} latest The most recent ranking of tracks.
+ * @param {Change[][]} changes The list of historical changes to the track
+ * ranking.
+ * @param {c3.ChartAPI} chart The chart to populate.
+ */
+function populateChart(latest, changes, chart) {
+// number of days we have backtracked
```
geekster777, line 137: nit: tab this comment in.

client/js/spotify-gdpr.js
```diff
@@ -0,0 +1,271 @@
...
+
+    // take the current ordering of the songs and store it in the historical
+    // index list
+    for (const [id, index] of indices.entries()) {
+      let indexHistory;
+      if (indicesHistory.has(id)) {
+        indexHistory = indicesHistory.get(id);
+        indexHistory.push(index);
+      } else {
+      // this song wasn't in the record, backfill all of the indices
```
geekster777, line 188: indent.

client/js/analysis/gdpr.js
```diff
@@ -0,0 +1,199 @@
...
+    zip.createReader(
+        new zip.BlobReader(data),
+        (reader) => {
+          reader.getEntries((files) => {
+            /** @typedef {{ index: number; file: zip.Entry; }} EntryWithIndex */
+
+            // get only entries named StreamingHistoryX.json and pair them with
+            // their index so they can be sorted
+            const entriesWithIndices =
+            files
```
geekster777, line 49: indent

client/js/analysis/gdpr.js
```diff
@@ -0,0 +1,199 @@
...
+                  }
+
+                  return arr;
+                }, [])
+                .sort((a, b) => a.index < b.index ? -1 : 1);
+
+            // create a Promise for each StreamingHistoryX.json to read & parse
+            // its content
+            const contentPromises =
+            entriesWithIndices.map(({index, file}) => {
```
geekster777, line 68: indent.

client/js/analysis/gdpr.js
```diff
@@ -0,0 +1,199 @@
...
+ * records.
+ *
+ * @param {Blob} data The zipped GDPR data.
+ * @returns {Promise<GDPRRecord[]>} A list of GDPR streaming records.
+ */
+export async function getStreamingData(data) {
+  const streamingDataPattern = /StreamingHistory(\d+)\.json$/i;
+
+  return new Promise((resolve, reject) => {
+    zip.createReader(
```
geekster777, line 40: The functions in this reader are quite difficult to tease out given the heavy nesting of inlined-functions. You should instead split these functions out into separate functions that you call:
```
export async function getStreamingData(data) {
  ...
  const fileDataToUtcEntry = (fileData) => {
    ...
  };
 
  const readFiles = (files) => {
    ...
    const contentPromises = entriesWithIndices.map(...).map(fileDataToUtcEntry);
    Promise.all(contentPromises).then(entries => resolve(entries));
  }

  zip.createReader(new zip.BlobReader(data), (reader) => reader.getEntries(readFiles));
```

PR #26, opened 2020-07-08
Youtube genre count

gracetian6:
**Displays frequency count of music genres of last 5 liked youtube videos** 
image

- reorganized `YoutubeServlet.java` with helper functions
- `getYoutubeRes()` is the same as before. It makes youtube api call and gets json string.
- `updateMusicCount()` is new. It updates a hash map with the frequency count in picture above.


server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -40,14 +40,78 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
+    protected void updateMusicCount(String youtubeResBody, HashMap<String, Integer> genreCount) {
+        JsonObject jObject = JsonParser.parseString(youtubeResBody).getAsJsonObject();
+        JsonArray videos = jObject.getAsJsonArray("items");
+
+        // iterates through each liked video
+        for (int i = 0; i < videos.size(); i++) {
+            // extracts array of topicCategories in video
+            JsonObject video = videos.get(i).getAsJsonObject();
+            JsonObject topicDetails = video.getAsJsonObject("topicDetails");
+            JsonArray topicCategories = topicDetails.getAsJsonArray("topicCategories");
```
gracetian6, line 79: Advice? I'm going from items > topicDetails > topicCategories in JSON object. This works if json response works as it should, but probably need to error check each step of the way.

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -40,14 +40,78 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
+
+    /**
+     * checks whether topic is categorized as music
+     * by checking if the last word is "music" or "Music"
+     * @param topic 
+     * @return Boolean
+     */
+    protected Boolean isMusic(String topic) {
+        String lastWord = topic.substring(topic.lastIndexOf(" ") + 1);
+        return lastWord.equals("Music") || lastWord.equals("music");
```
laptou, line 61: you can use `equalsIgnoreCase("music")`

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
laptou: well, the YouTube API is not likely to just change suddenly. since this is not user-generated data, you can probably get away with not error-checking it; even if an exception is thrown it just turns the request into a 500 error. the most Correct™️ thing to do though is to check each of these JSON objects before using them and return a 400 error if any of them are `null`.

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -40,14 +40,78 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
+        // retrieve Youtube API key and access token 
+        String API_KEY = Secrets.getSecretString("YOUTUBE_API_KEY");
+        HttpSession session = req.getSession();
+        var accessToken = session.getAttribute("oauth-access-token-youtube");
+        if (accessToken == null) {
+            res.setStatus(401);
+            return;
+        }
+        String youtubeResBody = getYoutubeRes(req, res, API_KEY, accessToken.toString());
+        HashMap<String, Integer> genreCount = new HashMap<String, Integer>();
```
laptou, line 113: nit: use `var` for variables where the type is obvious

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -40,14 +40,78 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
         var httpClient = HttpClient.newHttpClient();
         var youtubeReq = HttpRequest.newBuilder(youtubeUri)
-                .header("Authorization", "Bearer " + accessToken.toString())
+                .header("Authorization", "Bearer " + accessToken)
                 .header("Accept", "application/json")
                 .GET()
                 .build();
 
         // get response to Youtube API 
         var youtubeRes = httpClient.sendAsync(youtubeReq, BodyHandlers.ofString()).join();
```
laptou, line 49: nit: don't use `var` for variables where the type is not obvious

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,22 +15,22 @@
...
 @WebServlet("/api/youtube")
 public class YoutubeServlet extends HttpServlet {
-    @Override
-    protected void doGet(HttpServletRequest req, HttpServletResponse res) 
+    /**
+     * makes http request of youtube api, gets json string of youtube res
+     * @param req
+     * @param res
+     * @param API_KEY
+     * @param accessToken
```
laptou, line 27: `@param` tags should have a description of what each parameter does

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,22 +15,22 @@
...
+     * @return JSON string of youtube response
+     * @throws ServletException
+     * @throws IOException
+     */
+    protected String getYoutubeRes(HttpServletRequest req, HttpServletResponse res, String API_KEY, String accessToken)
```
laptou, line 32: nit: param names should be camelCase

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
gracetian6: oops haha, will fix

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
gracetian6: @ everyone else I'm going to deal with case where video does not have topicDetails json in an upcoming branch `youtube-merge-error`

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -2,7 +2,7 @@
 import com.google.musicanalysis.util.Secrets;
 import com.google.musicanalysis.util.URLEncodedBuilder;
 
-import com.google.gson.JsonParser;
+import com.google.gson.*;
```
seunomonije, line 5: nit: I personally only include the files that I need within a package. 

https://stackoverflow.com/questions/14897448/use-of-in-import-statement-in-java

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,39 +15,106 @@
...
-        HttpSession session = req.getSession();
-        var accessToken = session.getAttribute("oauth-access-token-youtube");
-        if (accessToken == null) {
-            res.setStatus(401);
-            return;
-        }
-    
         // make http request to youtube API
-        var youtubeParam = new URLEncodedBuilder()
+        URLEncodedBuilder youtubeParam = new URLEncodedBuilder()
```
seunomonije, line 34: Ah I see how URLEncodedBuilder works now. This is a cool way to do it bc it takes care of the encoding for you.

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
gracetian6: got it! I added 3 more packages

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,39 +17,106 @@
...
+     * parses through youtube liked videos json string,
+     * updates hash map to contain frequency count of each music genre
+     * @param youtubeResBody json response of youtube liked videos
+     * @param genreCount hash map of frequency count of each music genre
+     */
+    protected void updateMusicCount(String youtubeResBody, HashMap<String, Integer> genreCount) {
+        JsonObject jObject = JsonParser.parseString(youtubeResBody).getAsJsonObject();
+        JsonArray videos = jObject.getAsJsonArray("items");
+
+        // iterates through each liked video
```
jessicaslaughter, line 75: You don't need comments like these that just explain the mechanics of the code (in this case that the for loop is iterating over a list of videos). Reserve implementation comments for places where the intention of the code isn't immediately obvious.

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,39 +17,106 @@
...
+            JsonObject video = videos.get(i).getAsJsonObject();
+            JsonObject topicDetails = video.getAsJsonObject("topicDetails");
+            if (topicDetails == null) {
+                // skip this video if it doesn't have topic details
+                continue;
+            }
+            JsonArray topicCategories = topicDetails.getAsJsonArray("topicCategories");
+
+            for (int j = 0; j < topicCategories.size(); j++) {
+                // extract music genre out of wikipedia links of topic categories
```
jessicaslaughter, line 87: this comment is more helpful because it's not obvious that the link has a music genre in it

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,39 +17,106 @@
...
+                String link = topicCategories.get(j).toString();
+                String topic = link.substring(link.lastIndexOf('/') + 1);
+                topic = topic.replaceAll("\"", "");
+                topic = topic.replaceAll("_", " ");
+
+                // skip if topic is not music
```
jessicaslaughter, line 93: Also not a needed comment, since the code is self-explanatory

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,39 +17,106 @@
...
+                if (!isMusic(topic)) {
+                    break;
+                }
+
+                // update genreCount hashmap
```
jessicaslaughter, line 98: Same here, saying that you're adding something to a hashmap doesn't tell me anything about the intent of this code, just the mechanics

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,39 +17,106 @@
...
+                genreCount.put(topic, count + 1);
+            }
+        }
+        return;
+    }
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse res) 
+        throws ServletException, IOException {
+        // retrieve Youtube API key and access token
```
geekster777, line 105: unnecessary comment.

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,39 +17,106 @@
...
+        throws ServletException, IOException {
+        // retrieve Youtube API key and access token 
+        String API_KEY = Secrets.getSecretString("YOUTUBE_API_KEY");
+        HttpSession session = req.getSession();
+        var accessToken = session.getAttribute("oauth-access-token-youtube");
+        if (accessToken == null) {
+            res.setStatus(401);
+            return;
+        }
+        String youtubeResBody = getYoutubeRes(API_KEY, accessToken.toString());
```
geekster777, line 113: nit: newline before this line.

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,39 +17,106 @@
...
+        }
+        return;
+    }
+
+    @Override
+    protected void doGet(HttpServletRequest req, HttpServletResponse res) 
+        throws ServletException, IOException {
+        // retrieve Youtube API key and access token 
+        String API_KEY = Secrets.getSecretString("YOUTUBE_API_KEY");
+        HttpSession session = req.getSession();
```
geekster777, line 107: nit: add a newline before this code since this is a discrete chunk of getting our access token from our session.

PR #28, opened 2020-07-09
Youtube genres on front end

gracetian6:
- Instead of writing to hash map of genre count from backend, I moved the hash map to the front end on `youtube-genre.html` using `genre.js` to fetch from `/api/youtube`.
- `genre.js` also redirects if `/api/youtube` gives an error message.

image

**If any of you have better name suggestions, please let me know! I'm not sure if I really like the names `genre.js` and `youtube-genre.html`.**


client/js/genre.js
```diff
@@ -0,0 +1,17 @@
+/** fetches youtube api calls from YoutubeServlet.java
+    and displays on youtube-genre.html
+*/
+
+const genreBlock = document.getElementById('genres');
+async function displayMusicGenre() {
+  const response = await fetch("/api/youtube");
+  console.log(response.status);
```
seunomonije, line 8: nit: remove debug

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -15,39 +18,110 @@
...
+                // skip this video if it doesn't have topic details
+                continue;
+            }
+            JsonArray topicCategories = topicDetails.getAsJsonArray("topicCategories");
+
+            for (int j = 0; j < topicCategories.size(); j++) {
+                // extract music genre out of wikipedia links of topic categories
+                String link = topicCategories.get(j).toString();
+                String topic = link.substring(link.lastIndexOf('/') + 1);
+                topic = topic.replaceAll("\"", "");
```
seunomonije, line 92: I think you can do this with regex like this:
```topic = topic.replaceAll(["_], "")```
The brackets create a character group/class that regex will match in the string.  Every instance of either of these will be replaced with a space.

Comment (seunomonije) [reply server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java:92]: My regex string ```"[\"_]"``` might be wrong, but you get the general idea lol

Comment (gracetian6) [reply server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java:92]: oh nice I'll look into this

File: server/pom.xml [hunk 3]
```diff
@@ -38,6 +38,26 @@
         <artifactId>gson</artifactId>
         <version>2.8.6</version>
     </dependency>
+    <!-- Music Match -->
+    <dependency>
+      <groupId>com.sachinhandiekar</groupId>
+      <artifactId>jMusixMatch</artifactId>
+      <version>1.1.4</version>
+    </dependency>
```
laptou, line 46: get rid of this

gracetian6: got it

gracetian6:
fixed the lint errors I could find on my files. There are still more lint errors on other files I did not work with: 
image

PR #29, opened 2020-07-10
Donut charts

seunomonije:
Finished the donut charts in a scaleable way.  Each category will have a different color. I will also change the code to allow for more modification once everything is in master and I can actually run the front-end code on my computer.

JSFiddle: https://jsfiddle.net/ncdqm8hs/1/

And if that doesn't work, screen recording: https://drive.google.com/file/d/1zxy3eArrNZG_fTeJYt00xAc_gZFPBgKy/view?usp=sharing


seunomonije:
@gracetian6 I made it so both of yall are "editors" and can share it. Everyone else who has the link may comment.

client/css/analysis.css
```diff
@@ -0,0 +1,43 @@
+:root {
+    /* the higher the number, the less filled */
+  --percent: 8;
```
laptou, line 3: > the higher the number, the less filled 

Why did you do it like that? Also, why did you put the variable on `:root` instead of on the donut chart container?

You could make the `--percent` variable work like an actual percentage by doing this:
- set the `pathLength` attribute of the `circle` element to 100; then the browser will always do stroke calculations as if the circle's circumference is 100 units (which means a stroke dash offset of 1 will actually offset the stroke by 1%)
- use `calc()` to do subtraction at the site where you use the variable:
  ```css
  stroke-dashoffset: calc(100 - var(--percent));
  ```

client/css/analysis.css
```diff
@@ -0,0 +1,43 @@
...
+  to {
+    stroke-dashoffset: var(--percent); /* half filled is 220*/
+  }
+}
+
+@keyframes donut {
+  to {
+    stroke-dashoffset: var(--percent);
+  }
+}
```
laptou, line 43: If you do the animation using `@keyframes`, then it will not be very easy for you to control the percentage from JS. Maybe use the `transition` property instead?

client/js/analysis.js
```diff
@@ -1,27 +1,86 @@
-
-
+// Will be overwritten in merge
 window.onload = function() {
```
laptou, line 2: nit: prefer `addEventListener('onload', () => ...)` to `window.onload = function() { ... }`

client/css/analysis.css
seunomonije: Why would it not be easy to control if I just change the variable name for each chart?

client/css/analysis.css
seunomonije: 1. I put the variable on root because there is no container in this pr to add it to.
2. I did it like this so that in JS, I could easily change the `--percent` variable for each donut from there and it would fill the circle accordingly. 440 is approximately 2πr where r is 70 as seen in the code.
   - For example, if Perspective returns 70% toxicity, do 440 - 440*.70 in js and then set the `--percent` variable from there. Very simple. 
3. I tried to set the `pathLength` attribute to 100 and try it that way, but changing the `stroke-dashoffset` had no effect.

client/css/analysis.css
laptou: `@keyframes` won't automatically re-run when you change the value of the `--percent` variable, but a transition will

seunomonije: The charts won't ever have to re-run. The page will auto reload for every new song and will create new charts with values.

client/css/analysis.css
laptou: Check this out: https://jsfiddle.net/jc5pn9rm/

laptou: And you can use `calc()` if you want the donut to fill up counterclockwise instead of clockwise

seunomonije: Very very nice. Thank you

client/css/analysis.css
seunomonije: Another issue with not using keyframes is that the animation doesn't start from 0. I want it to be able to start empty and then fill and its easier to use keyframes with that rather than having to js functions @laptou

seunomonije:
Will include these in better pull request.

PR #30, opened 2020-07-10
send # of vids from frontend to backend

gracetian6:
- added input form that allows users to enter number of videos desired as variable `numVideos`
- sends `numVideos` from front end to backend by updating as `?num_videos=` in URL
- `numVideos` is a dummy variable in the backend. My next step is to get that working in backend.

image

- EDIT: deleted form POST and changed to `onclick=js function` instead


client/js/genre.js
```diff
@@ -1,11 +1,28 @@
...
+/**
+ * fetches genre count hashmap from /api/youtube and updates html
+ */
 async function displayMusicGenre() {
-  const response = await fetch("/api/youtube");
-  console.log(response.status);
+  const genreBlock = document.getElementById('genres');
+  const numVideos = getNumVids();
+  const response = await fetch(`/api/youtube?num_videos=${numVideos}`);
   if (response.status != 200) {
```
laptou, line 26: check if the response is `401`, not just if it's not `200`; all of the response codes have distinct meanings and most of them don't mean "you are not logged in / authorized"

gracetian6:
- deleted form POST and added js function instead. Gets rid of unnecessary Steps.

Review this PR on https://github.com/googleinterns/step99-2020/pull/28 before this one! (@ibiyemi)

gracetian6:
> LGTM. In the future, consider adding validation to your input so that the user can't enter that they want `onety-one` videos.

@laptou The html tag of my input is set to `type: number` so that words can't be submitted in the input?

client/js/genre.js
```diff
@@ -1,14 +1,32 @@
...
+const DEFAULT_NUM_VIDS = 10;
+
+/**
+ * extracts num vids from input, otherwise returns default
+ * helper fn for displayMusicGenre()
+ * @returns {number} num videos user wants to retrieve
+ */
+function getNumVids() {
+  const numVideos = numVideosInput.value;
+  return numVideos || DEFAULT_NUM_VIDS;
```
geekster777, line 17: nit: just return `numVideosInput.value || DEFAULT_NUM_VIDS`. Creating an intermediary const doesn't improve readability at all so it's not necessary.

PR #32, opened 2020-07-13
Create GDPR data chart

laptou:
- Line chart of GDPR data now works

Screenshot 2020-07-13 at 12 16 22 - Display 1


client/js/spotify-gdpr.js
```diff
@@ -3,214 +3,106 @@
...
-          const {oldItem, newItem} = change;
-
-          const idx = indices.get(newItem);
-          current[idx] = oldItem;
-
-          indices.set(newItem, null);
-          indices.set(oldItem, idx);
-          break;
-        }
+async function populateChart(collatedRecords, chart) {
```
seunomonije, line 25: I see what you're doing here

client/js/spotify-gdpr.js
```diff
@@ -3,214 +3,106 @@
...
-    generation++;
+    // turn into array of entries
+    [...rollingSums]
+        // sort by play time in the last 28 days
+        .sort((entryA, entryB) => {
+          if (entryA[1] > entryB[1]) return -1;
+          if (entryA[1] < entryB[1]) return 1;
+          return 0;
+        })
+        // select top 50
```
jessicaslaughter, line 67: should this be 15?

client/js/spotify-gdpr.js
```diff
@@ -3,214 +3,106 @@
...
-    [],
-    [],
-    [],
-  ];
-}
-
-const MS_PER_DAY = 24 * 60 * 60 * 1000;
 
 /**
  * Populates `chart` with historical track data derived from `latest` and
```
jessicaslaughter, line 18: This comment should be updated

client/js/spotify-gdpr.js
```diff
@@ -3,214 +3,106 @@
...
+    // turn into array of entries
+    [...rollingSums]
+        // sort by play time in the last 28 days
+        .sort((entryA, entryB) => {
+          if (entryA[1] > entryB[1]) return -1;
+          if (entryA[1] < entryB[1]) return 1;
+          return 0;
+        })
+        // select top 50
+        .slice(0, 15)
```
geekster777, line 68: top 50 or 15? One of these numbers is wrong.

client/js/spotify-gdpr.js
```diff
@@ -3,214 +3,106 @@
...
+  for (let i = 0; i < chartData.length; i++) {
+    const chartDataSegment = chartData.slice(i, i + 15);
+    await new Promise((resolve) => chart.load({
+      columns: [
+        ['date', ...rankingDates],
+        ...chartDataSegment,
+      ],
+      done: resolve,
+    }));
+    await sleep(400);
```
geekster777, line 104: Why the sleep, and why 400ms?

client/js/spotify-gdpr.js
geekster777: This number should probably be in a const.

client/js/spotify-gdpr.js
laptou: > `// load data in segments to avoid hanging the browser`

JS execution is single-threaded, so there's no way for me to do such an expensive operation without causing the window to hang for tens of seconds. The `sleep` yields control to the browser temporarily to make things remain responsive.

400ms is the Doherty Threshold, but I could've picked any number less than 500 I suppose, even 0.

geekster777: > 400ms is the Doherty Threshold
That was a fun read.

Have you experienced hanging when you don't do this? I'm pretty sure you're already yielding the rendering thread when you `await chart.load(...)` since your zip library uses webworkers (iirc). If you do experience hanging, can you add a comment that the 15 segments is meant to be roughly 400ms worth of work, so you're doing 400ms on and 400ms off. Otherwise it seems like the 400ms is there to allow you to wait on some specific task to be completed.

laptou: The zip library uses web workers, but at this point, the zip library is no longer involved (the data has already been loaded and is now being stuffed into the chart). I did experience extreme hanging when I didn't do this, to the point where it was kind of unusable.

geekster777: SGTM. Yeah, just document roughly how long the chunks hang, and try get that number under 400ms.

PR #33, opened 2020-07-13
Use Google charts

laptou:
- C3.js charts are just too slow to display this amount of data, so Google Charts is used instead

Screenshot 2020-07-13 at 12 18 04 - Display 1


laptou:
> Didn't you mention that you tried c3.js and google charts, but decided to make your own custom chart instead? Is this in a future PR?

Yep

client/js/spotify-gdpr.js
```diff
@@ -82,82 +82,71 @@ async function populateChart(collatedRecords, chart) {
...
-          trackHistory.push(index);
+          trackHistory.push(index + 1);
         });
   }
 
-  // format of a chart series in c3 is ['name of series', point1, point2, ...]
-  const chartData = Array
-      .from(rankingHistory.entries())
-      .map(([id, history]) => [id, ...history]);
+  const series = Array.from(rankingHistory.entries());
```
jessicaslaughter, line 89: nit: consider renaming to chartSeries for clarity

PR #35, opened 2020-07-14
Create custom chart

laptou:
- C3.js was too slow
- Google Charts didn't have enough flexibility to do what I was trying to do and look good at the same time
- This PR introduces the beginnings of a custom chart that uses SVG

Screenshot 2020-07-13 at 13 39 30 - Display 1


client/js/chart/index.js
```diff
@@ -0,0 +1,130 @@
...
+  svg.append(createDefs());
+
+  const seriesContainer = document.createElementNS(SVG_NS, 'g');
+  seriesContainer.setAttribute('class', 'series-container');
+  svg.append(seriesContainer);
+
+  let index = 0;
+  const colors = ['blue', 'red', 'green', 'purple', 'orange'];
+
+  for (const [, history] of [...rankingHistory].slice(0, 7)) {
```
seunomonije, line 23: A comment here would be nice.

client/js/chart/index.js
```diff
@@ -0,0 +1,130 @@
...
+  endCap.setAttribute('class', 'series-run-cap');
+
+  startCap.setAttribute('r', '5');
+  startCap.setAttribute('cx', start * RUN_SCALE_X + 'px');
+  startCap.setAttribute('cy', history[start] * RUN_SCALE_Y + 'px');
+  endCap.setAttribute('r', '5');
+  endCap.setAttribute('cx', (end - 1) * RUN_SCALE_X + 'px');
+  endCap.setAttribute('cy', history[end - 1] * RUN_SCALE_Y + 'px');
+
+  return [startCap, line, touchTarget, endCap];
```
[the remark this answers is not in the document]
laptou: Can you elaborate please? I don't really get what is difficult to understand about this.

client/js/chart/index.js
```diff
@@ -0,0 +1,131 @@
...
+
+  const seriesContainer = document.createElementNS(SVG_NS, 'g');
+  seriesContainer.setAttribute('class', 'series-container');
+  svg.append(seriesContainer);
+
+  let index = 0;
+  const colors = ['blue', 'red', 'green', 'purple', 'orange'];
+
+  // get only the first 7 tracks for now
+  for (const history of [...rankingHistory.values()].slice(0, 7)) {
```
gracetian6, line 24: What is the `...` notation mean in `[...rankingHistory]`?

client/js/chart/index.js
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Creates a new series (set of lines on the chart for a specific song).
+ *
+ * @param {string} color The color of this series.
+ * @param {number[]} history The historical positions of this track on the
+ * leaderboard.
+ * @returns {SVGGElement} A group containing the series.
+ */
+function createSeries(color, history) {
+  const series = document.createElementNS(SVG_NS, 'g');
```
gracetian6, line 46: helpful comments!

client/js/chart/index.js
```diff
@@ -0,0 +1,131 @@
...
+
+  startCap.setAttribute('class', 'series-run-cap');
+  endCap.setAttribute('class', 'series-run-cap');
+
+  startCap.setAttribute('r', '5');
+  startCap.setAttribute('cx', start * RUN_SCALE_X + 'px');
+  startCap.setAttribute('cy', history[start] * RUN_SCALE_Y + 'px');
+  endCap.setAttribute('r', '5');
+  endCap.setAttribute('cx', (end - 1) * RUN_SCALE_X + 'px');
+  endCap.setAttribute('cy', history[end - 1] * RUN_SCALE_Y + 'px');
```
gracetian6, line 103: can you explain what svg elements are? (Would like to know for future reference cause my data visualization w/ d3.js uses `svg`s)

Also, can you explain the `setAttribute`s? From the code it looks like you're setting the radius (r?) and x/y positions

client/js/chart/index.js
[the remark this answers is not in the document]
gracetian6: I think seun is saying it makes more sense to call an object attribute/method to get `startCap`, `line`, `touchTarget`, `endCap` respectively, instead of getting indices 0-3 of this list. With a list, if you want `startCap` you'd call `list[0]`. This seems kind of like a magic number instead of calling a method/attribute `.getStartCap()` of the object

But since you're iterating through this list, on line 58, returning an object doesn't make sense.

client/css/spotify-gdpr-chart.css
```diff
@@ -1,76 +1,30 @@
...
+.chart {
+  width: 100%;
+  height: 40rem;
 }
 
-.c3 text {
-  fill: var(--color-foreground);
+.chart .series {
+  filter: url('#filter-highlight');
 }
```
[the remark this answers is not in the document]
laptou: [`filter`](https://developer.mozilla.org/en-US/docs/Web/CSS/filter) applies visual effects such as saturation, shadows, blurring, etc

client/js/chart/index.js
laptou: [Spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax)

client/js/chart/index.js
laptou: Yep. MDN has great documentation of all of the SVG elements and their attributes, including [`<circle>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/circle)

client/js/chart/index.js
```diff
@@ -0,0 +1,131 @@
...
+    while (start < history.length && history[start] === null) {
+      start++;
+    }
+    end = start + 1;
+    while (end < history.length && history[end] !== null) {
+      end++;
+    }
+    series.append(...createRun(history, start, end));
+    start = end;
+  }
```
geekster777, line 61: This loop is a bit difficult to follow. Consider splitting up the calls to createRun and instead find all the run segments first, then create the runs:
```
const segments = [];
while (end < history.length) {
  // Null history nodes represent the end of a segment.
  if (history[end] !== null) {
    end++;
    continue;
  }

  if (end - start) > 0) {
    segments.push({start, end});
  }
  end++;
  start = end;
}

series.append(...segments.map(createRun));
```

client/js/chart/index.js
```diff
@@ -0,0 +1,131 @@
...
+      .slice(start, end)
+      .map((val, idx) => `${(idx + start) * RUN_SCALE_X},${val * RUN_SCALE_Y}`)
+      .join(' ');
+
+  // line that is displayed
+  const line = document.createElementNS(SVG_NS, 'polyline');
+  line.setAttribute('class', 'series-run');
+  line.setAttribute('points', points);
+
+  // secondary invisible line to make it easier to hit the line
```
geekster777, line 86: nit: mention that this line is wider, which is what makes it easier to touch.

client/js/chart/index.js
```diff
@@ -0,0 +1,131 @@
...
+ * Creates a new run (continuous group of points within a series).
+ *
+ * @param {number[]} history The historical positions of this track on the
+ * leaderboard.
+ * @param {number} start The index of the first entry in this run.
+ * @param {number} end The index of the last entry in this run.
+ * @returns {SVGElement[]} The elements that compose this run.
+ */
+function createRun(history, start, end) {
+  const points = history
```
geekster777, line 76: It'll be more readable if you separate this out into 2 variables: points and pointsStr. The points are just the x/y pairs for each of these points, whereas pointsStr is the concatedated pair in "x,y x,y..." format. This will help when you create the end caps since the code becomes:
```
startCap.setAttribute('cx', points[0].x + 'px');
startCap.setAttribute('cy', points[0].y + 'px');
```

client/js/chart/index.js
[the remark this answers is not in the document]
geekster777: Can you instead wrap up each run in its own element and just return that? This will also make your svg easier to read/debug:
```
const container = document.createElementNS(SVG_NS, 'g');
container.append(startCap, line, touchTarget, endCap);
return container;
```

client/js/chart/index.js
laptou: I don't really think that makes it clearer. I just added more comments. Does that work?

client/js/chart/index.js
```diff
@@ -0,0 +1,141 @@
+import {SVG_NS} from '../util.js';
+
+/**
+ * Creates an SVG chart inside of `el` with the given data.
```
jessicaslaughter, line 4: nit: update 'el'

client/js/chart/index.js
```diff
@@ -0,0 +1,141 @@
...
+function createSeries(history, color) {
+  const series = document.createElementNS(SVG_NS, 'g');
+  series.setAttribute('class', 'series');
+  series.style.setProperty('--run-color', color);
+
+  let start = 0;
+  let end = 0;
+
+  while (end < history.length) {
+    // go until we find a non-null point
```
jessicaslaughter, line 54: These comments would be more helpful if they described why we're looking for null and non-null points. Like Nick pointed out, it's important to indicate that null points represent the end of a segment or are otherwise not in a segment. So this particular comment should be something like, "The first non-null history node represents the beginning of a segment."

client/js/chart/index.js
geekster777: It's clearer because it's a single while loop rather than 2 nested while loops. Please fix.

laptou: Done

PR #41, opened 2020-07-14
Tooltip performance improvements

laptou:
- It turns out that putting an SVG filter on every line in the chart is terrible for performance
- This PR avoids doing that


client/js/gdpr/chart/index.js
```diff
@@ -244,27 +226,22 @@ function createDefs() {
...
-  const dimFilter = document.createElementNS(SVG_NS, 'filter');
-  dimFilter.id = 'filter-dim';
-  dimFilter.innerHTML = `
-    <feComponentTransfer>
-      <feFuncR type="linear" slope="0.5" />
-      <feFuncG type="linear" slope="0.5" />
-      <feFuncB type="linear" slope="0.5" />
-    </feComponentTransfer>
-  `;
-
```
gracetian6, line 256: so this gets rid of SVG filter?

client/js/gdpr/chart/index.js
```diff
@@ -116,35 +127,6 @@ export function createChart(el, rankingHistory, rankingDates) {
...
-  // value in history
-  const y = Math.round(chartPos.y / RUN_SCALE_Y);
-
-  const hit = rows[x].findIndex((rank) => rank === y);
-
-  if (hit < 0) return {hit: null, x: null, y: null};
-
-  return {hit, x, y};
-}
-
```
gracetian6, line 147: is this deleted just because most of it is added above in lines 59-68? What's your reason for deleting it? 

Your `svg.addEventListener('mousemove', ({clientX, clientY}) => {` is getting kind of long so I would avoid adding more lines

client/js/gdpr/chart/index.js
laptou: yes?

client/js/gdpr/chart/index.js
laptou: The function felt clunky and prevented me from doing [this optimization](https://github.com/googleinterns/step99-2020/blob/761d86798a14dee9ac2474c85fcd6af9a2a17651/client/js/chart/index.js#L63), so I scrapped it.

PR #43, opened 2020-07-14
Frontend Scroller (Youtube Genres)

gracetian6:
See [DEMO here](https://drive.google.com/file/d/1ZrsnZw86tccwCfx5MeD0A0wEnxO259E_/view?usp=sharing)

- add hero title page 
- added scroller with two columns. The first column will contain text (that scrolls), the second data column will contain data visualization (which is sticky)
- added d3.js package


client/youtube-genre.html
```diff
@@ -4,19 +4,42 @@
...
+            Section 5
+          </section>
+        </div>
+        <div id='vis'>
+          add visualizations here
+        </div>
+      </div>
+    </main>
   </body>
+  <script src="https://d3js.org/d3.v4.min.js"></script>
```
laptou, line 43: Why are you using v4 instead of v5?

gracetian6: just because the d3 examples I'm looking at are using v4 😟 

I'll change to v5 but worried about deprecated functions

client/css/scroll.css
```diff
@@ -0,0 +1,44 @@
...
+  display: grid;
+  align-content: center;
+  justify-items: center;
+  width: 100%; /* don't use 100vw, that will cause horizontal scrollbar */
+  height: 100vh;
+  
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: cover;
+  background-attachment: fixed;
```
geekster777, line 38: You have no background here. Why include this css?

client/css/scroll.css
```diff
@@ -0,0 +1,44 @@
...
+  width: 50vw;
+
+  height: 60vh;
+  margin-left: 0;
+
+  background-color: #dddddd;
+}
+
+.hero {
+  display: grid;
```
geekster777, line 29: Why display grid? Use `display: flex;` since it's less tightly coupled with a grid layout and will behave closer to a regular block element.

geekster777: http://howtocenterincss.com/ is a good resource for determining how to vertically center elements.

gracetian6: When I followed `display:flex` and the link, the divs didn't center. The text went on left and right sides.

PR #42, opened 2020-07-14
Add gridlines for dates

laptou:
- Add gridlines for dates

Screenshot 2020-07-14 at 15 11 22 - Display 1

PR #45, opened 2020-07-15
frontend integration

seunomonije:
The front-end MVP code for the website.

Here is the link to a screencast: https://drive.google.com/file/d/1wGbeZvgFbaH8CD0aL3fmpyXwBv7Lv8FJ/view?usp=sharing

### DIFF
--- a/client/analysis.html
+++ b/client/analysis.html
@@ -11,7 +11,7 @@
 
   <body>
     <div id="content">
-        <div class = "flex-row center">
+        <div class="flex-row center">
             <div id="header-wrapper">
                 <h1>Enter video name below:</h1>
             </div>
--- a/client/analysis.html
+++ b/client/analysis.html
@@ -4,20 +4,40 @@
   <head>
     <meta charset="utf-8">
     <meta name="description" content="STEP Capstone">
+    <link rel="stylesheet" href="css/base.css">
     <link rel="stylesheet" href="css/analysis.css">
     <title>Song Analysis</title>
   </head>
 
   <body>
-    <h1>Enter a song name and artist name</h1>
+    <div id="content">
+        <div class = "flex-row center">
+            <div id="header-wrapper">
+                <h1>Enter video name below:</h1>
+            </div>
+        </div>
 
-    <label for="video-name">Insert Video name</label>
-    <input type="text" placeholder="Video name" name="video-name"/>
+        <div class="flex-row center">
+            <div id="search-wrapper">
+                <input type="text" id="searchbar" />
+                <button id="sendbutton">Send</button>
+            </div>
+        </div>
 
-    <input type="submit" value="Submit" id="formSubmit" />
+        <div class="flex-row center">
+            <p></p>
+        </div>
 
-    <h1>Response</h1>
-    <div id="response"></div>
+        <div class="flex-row" id="middle-content">
+            <div class="flex-column">
+                <h2>Featured comments:</h2>
+                <ol id="list"></ol>
+            </div>
+            <div class="flex-column">
+                <div class="grid-container" id="charts"></div>
+            </div>
+        </div>
+    </div>
   </body>
   <script src="js/analysis.js"></script>
 </html>
--- a/client/css/analysis.css
+++ b/client/css/analysis.css
@@ -106,8 +106,6 @@ input:focus, button:focus {
   padding: 5px;
 
   color: black;
-
-  
   font-size: 30px; /*light-gray*/
 
   background-color: #d3d3d3;
@@ -163,12 +161,12 @@ svg {
   transform: rotate(-90deg); 
 }
 
-.circle_animation {
+.circle-animation {
   stroke-dasharray: 100;
   stroke-dashoffset: 100;
 }
 
-.donut .circle_animation {
+.donut .circle-animation {
   -webkit-animation: donut 1.3s ease-in-out forwards;
   animation: donut 1.3s ease-in-out forwards;
 }
--- a/client/css/analysis.css
+++ b/client/css/analysis.css
@@ -1,49 +1,56 @@
 h2 {
-    padding: 15px;
+  padding: 15px;
 }
 
 ol {
   list-style: none;
+
   counter-reset: item;
 }
 
 li {
-  counter-increment: item;
   margin-bottom: 5%;
-  animation: fadeIn 1s;
+
   font-size: 20px;
-  font-family: Helvetica;
+  font-family: 'Helvetica', sans-serif;
+
+  animation: fade-in 1s;
+
+  counter-increment: item;
 }
 
 li:before {
-  text-align: center;
   display: inline-block;
-  content: counter(item);
+  width: 1.2em;
   margin-right: 10px;
+  padding: 5px;
+
+  color: white;
+  text-align: center;
+
   background-color: #9f86c0;
   border-radius: 100%;
-  padding: 5px;
-  color: white;
-  width: 1.2em;
+
   animation: pulse .7s ease-in;
+
+  content: counter(item);
 }
 
-input:focus,
-button:focus {
+input:focus, button:focus {
   outline: none;
 }
 
 @keyframes pulse {
   0% {
-    box-shadow: 0 0 0 0 rgba(0, 0, 0, 0.7);
+    box-shadow: 0 0 0 0 rgb(0 0 0 / 0.7);
   }
 
   100% {
-    box-shadow: 0 0 0 10px rgba(0, 0, 0, 0);
+    box-shadow: 0 0 0 10px rgb(0 0 0 / 0);
   }
 }
 
-@keyframes fadeIn {
+@keyframes fade-in {
   0% {
     opacity: 0;
   }
@@ -54,7 +61,7 @@ button:focus {
 }
 
 .fade {
-  animation: fadeIn .5s;
+  animation: fade-in .5s;
 }
 
 .flex-row {
@@ -65,9 +72,9 @@ button:focus {
 
 .flex-column {
   display: flex;
-  flex-direction: column;
+  flex: 1;
   flex-basis: 100%;
-  flex: 1;  
+  flex-direction: column;  
 }
 
 .center {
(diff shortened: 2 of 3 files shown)


gracetian6:
make link sharing for screen cast public! 😿

client/css/analysis.css
```diff
@@ -0,0 +1,171 @@
...
+  counter-reset: item;
+}
+
+li {
+  counter-increment: item;
+  margin-bottom: 5%;
+  animation: fadeIn 1s;
+  font-size: 20px;
+  font-family: Helvetica;
+}
```
gracetian6, line 16: stylelint fixes:
```
li {
  margin-bottom: 5%;

  font-size: 20px;
  font-family: Helvetica;

  animation: fadeIn 1s;

  counter-increment: item;
}
```

client/css/analysis.css
```diff
@@ -0,0 +1,171 @@
+h2 {
+    padding: 15px;
+}
+
+ol {
+  list-style: none;
+  counter-reset: item;
+}
```
gracetian6, line 8: stylelint fix:

```
ol {
  list-style: none;

  counter-reset: item;
}
```

client/css/analysis.css
```diff
@@ -0,0 +1,171 @@
...
+  padding: 5px;
+  color: white;
+  width: 1.2em;
+  animation: pulse .7s ease-in;
+}
+
+input:focus,
+button:focus {
+  outline: none;
+}
```
gracetian6, line 34:
```
input:focus, button:focus {
  outline: none;
}
```

client/css/analysis.css
```diff
@@ -0,0 +1,171 @@
...
+    margin: 10vh auto;
+}
+
+#header-wrapper {
+    width: 100%;
+}
+
+#search-wrapper {
+    width: 100%;
+}
```
gracetian6, line 95: indent should be like
```
#header-wrapper {
  width: 100%;
}

#search-wrapper {
  width: 100%;
}
```

client/css/analysis.css
```diff
@@ -0,0 +1,171 @@
...
+#searchbar {
+  border-radius: 5px;
+  border: none;
+  background-color: #d3d3d3;
+  /*light-gray*/
+  font-size: 30px;
+  padding: 5px;
+  color: black;
+  width: 40%;
+}
```
gracetian6, line 106:
```
#searchbar {
  width: 40%;
  padding: 5px;

  color: black;

  font-size: 30px;

  background-color: #d3d3d3; /*light-gray*/
  border: none;
  border-radius: 5px;
}
```

client/css/analysis.css
```diff
@@ -0,0 +1,171 @@
...
+#sendbutton {
+  font-size: 30px;
+  font-weight: bold;
+  background-color: #9f86c0;
+  border: none;
+  color: white;
+  border-radius: 5px;
+  text-align: center;
+  padding: 5px 10px;
+}
```
gracetian6, line 117:
```
#sendbutton {
  padding: 5px 10px;

  color: white;
  font-weight: bold;
  font-size: 30px;
  text-align: center;

  background-color: #9f86c0;
  border: none;
  border-radius: 5px;
}
```

client/css/analysis.css
```diff
@@ -0,0 +1,171 @@
...
+
+#overall {
+  margin-top: 10%;
+  background-color: #ff1493;
+  padding: 10px;
+  width: 80%;
+  border-radius: 5px;
+}
```
gracetian6, line 125:
```
#overall {
  width: 80%;
  margin-top: 10%;
  padding: 10px;

  background-color: #ff1493;
  border-radius: 5px;
}
```

client/css/analysis.css
```diff
@@ -0,0 +1,171 @@
...
+
+.item {
+  display: block;
+}
+
+.item h2 {
+  text-align: center;
+  width: 100%;
+}
```
gracetian6, line 139:
```
.item h2 {
  width: 100%;

  text-align: center;
}
```

client/css/analysis.css
```diff
@@ -0,0 +1,171 @@
...
+
+svg {
+  -webkit-transform: rotate(-90deg);
+  transform: rotate(-90deg);
+  display: block;
+  /* svg is auto display-inline*/
+  margin: auto;
+  overflow: visible; 
+}
```
gracetian6, line 148:
```
svg {
  display: block;

  /* svg is auto display-inline*/
  margin: auto;
  overflow: visible;

  -webkit-transform: rotate(-90deg);
  transform: rotate(-90deg); 
}
```

client/js/analysis.js
```diff
@@ -1,15 +1,123 @@
...
+  for (let i = 0; i < 6; i++) {
+    const values = [
+      'TOXICITY',
+      'IDENTITY ATTACK',
+      'INSULT',
+      'PROFANITY',
+      'THREAT',
+      'FLIRTATION',
+    ];
+    addDonutChart(values[i], Math.floor(Math.random() * 61) + 30);
```
gracetian6, line 20: get rid of magic nums 6, 1600, 61, 30, 9000

client/js/analysis.js
```diff
@@ -1,15 +1,123 @@
...
+  }
+};
+
+/**
+ * Adds an bullet point to the comment list
```
gracetian6, line 25: nit: change `an` to `a`

client/js/analysis.js
```diff
@@ -1,15 +1,123 @@
...
+
+  svg.setAttribute('xmlns', 'http://www.w3.org/2000/svg');
+
+  const g = buildCircle();
+  svg.appendChild(g);
+
+  div.appendChild(svg);
+
+  const currentDiv = document.getElementById('charts');
+  currentDiv.appendChild(div);
```
gracetian6, line 80: too many spaces. reorganize for readability. Each block should be discrete. (readjust as needed)
```
  const header = document.createElement('h2');
  header.innerText = str;
  div.appendChild(header);

  const svg = createSVGElement('svg');
  svg.setAttribute('xmlns', 'http://www.w3.org/2000/svg');

  const g = buildCircle();
  svg.appendChild(g);

  div.appendChild(svg);

  const currentDiv = document.getElementById('charts');
  currentDiv.appendChild(div);
```

client/js/analysis.js
```diff
@@ -1,15 +1,123 @@
...
+  const outsideCircle = createSVGElement('circle');
+
+  outsideCircle.setAttribute('class', 'circle_animation');
+  outsideCircle.setAttribute('r', '70');
+  outsideCircle.setAttribute('cx', '50%');
+  outsideCircle.setAttribute('cy', '50%');
+  outsideCircle.setAttribute('stroke-width', '25');
+  outsideCircle.setAttribute('stroke', '#ff1493'); // ff073a is good too
+  outsideCircle.setAttribute('fill', 'none');
+  outsideCircle.setAttribute('pathLength', '100');
```
gracetian6, line 107: might be able to avoid setting so many attributes at once by adding properties in css file like how ibiyemi did here for `verticalLine` svg:

- [css lines 46-51](https://github.com/googleinterns/step99-2020/pull/42/files#diff-df3e422d362ce99ee9375a71adbddf44R46-R51)
- [js lines 260-266](https://github.com/googleinterns/step99-2020/pull/42/files#diff-349c8b5f8ae66be1a9cfbe2f7817691dR260-R266)

client/js/analysis.js
seunomonije: 1500, 9000 aren't magic numbers. They are the number of milliseconds of delay before displaying each object. These must stay for the website to work: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout

6, 61, and 30 are there so that the code runs and you can see the animations. Wouldn't be able to have the screencast without them. Later today when I put everything together these will be replaced. 

6: The for loop starts at 1 and ends at 6. As you can see in line 5, I'm multiplying that number by 1500 to cause i*1500 ms of delay between each list element. This allows each list to appear staggered. When I put everything together, this will be replaced by something like `commentlist.length`

61 and 30 just set the range of the number `Math.random()` can generate. This generates a random percentage for each donut chart, which allows for the screencast and gives the effect that data is being handled in the donut charts even though it isn't at the moment. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random

client/js/analysis.js
seunomonije: The only things I could bring to css would be the stroke-width and fill, but r, cx, cy, pathLength, and stroke all need to be in js

seunomonije: I would prefer to keep it this way

gracetian6: sounds good

client/js/analysis.js
gracetian6: okay! As long as you replace hard coded nums later today (as you said) that sounds good.

client/css/analysis.css
```diff
@@ -0,0 +1,186 @@
...
+  list-style: none;
+
+  counter-reset: item;
+}
+
+li {
+  margin-bottom: 5%;
+
+  font-size: 20px;
+  font-family: 'Helvetica', sans-serif;
```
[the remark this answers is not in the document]
seunomonije: Because i want it to be helvetica

client/js/analysis.js
geekster777: For the random data, make it its own line with a comment:
```
// Mock the data with a number between 30% and 60% for now.
const chartFullness = Math.floor(Math.random() * 61 + 30);
addDonutChart(values[i], chartFullness);
```

client/js/analysis.js
```diff
@@ -1,15 +1,121 @@
...
+  insideCircle.setAttribute('fill', 'none');
+
+  const outsideCircle = createSVGElement('circle');
+
+  outsideCircle.setAttribute('class', 'circle-animation');
+  outsideCircle.setAttribute('r', '70');
+  outsideCircle.setAttribute('cx', '50%');
+  outsideCircle.setAttribute('cy', '50%');
+  outsideCircle.setAttribute('stroke-width', '25');
+  outsideCircle.setAttribute('stroke', '#ff1493'); // ff073a is good too
```
geekster777, line 103: Remove this comment

seunomonije:
Sorry @gracetian6 accidentally removed you

client/js/analysis.js
```diff
@@ -1,15 +1,127 @@
...
+ * Adds the feedback result element to the DOM tree
+ *
+ * @param {string} result Positive, Negative, or Neutral
+ */
+function addFeedbackResult(result) {
+  const el = document.createElement('div');
+  el.classList = 'center fade';
+  const text = document.createElement('h1');
+  text.setAttribute('id', 'overall');
+  text.innerHTML = 'Overall Response\n: ' + result;
```
geekster777, line 51: nit: insert a `<br>` instead of a `\n`

client/js/analysis.js
```diff
@@ -1,15 +1,127 @@
...
+/**
+ * Adds a donut chart
+ *
+ * @param {string} str the header string of the circle
+ * @param {string} percent the amount the circle will be filled
+ */
+function addDonutChart(str, percent) {
+  const div = document.createElement('div');
+
+  div.setAttribute('id', 'a-chart');
```
geekster777, line 67: Use a class instead of an id. Id's should be unique on your page, and you create 6 of these charts.

client/analysis.html
```diff
@@ -4,20 +4,40 @@
...
     <meta name="description" content="STEP Capstone">
+    <link rel="stylesheet" href="css/base.css">
     <link rel="stylesheet" href="css/analysis.css">
     <title>Song Analysis</title>
   </head>
 
   <body>
-    <h1>Enter a song name and artist name</h1>
+    <div id="content">
+        <div class = "flex-row center">
```
jessicaslaughter, line 14: nit: remove extra spaces around =

client/js/analysis.js
seunomonije: This was actually something I forgot to remove lol

client/js/analysis.js
```diff
@@ -64,7 +64,7 @@ function addFeedbackResult(result) {
 function addDonutChart(str, percent) {
   const div = document.createElement('div');
 
-  div.setAttribute('id', 'a-chart');
+  div.classList = 'a-chart';
   div.className = 'item donut';
```
jessicaslaughter, line 68: Using div.className will clear any classes applied to the div, so a-chart will no longer exist in its classList. You should add a-chart along with item and donut in the className assignment (classList.add would be used if you wanted to add these classes without affecting any other classes that might be applied to the div already).

PR #46, opened 2020-07-15
Reformat JSON data from backend

gracetian6:
Reformatted JSON response of the form 
```{"Music":6, "Pop music":2, "Rock music":1, "Country music":1, "Electronic music":1, "totalLiked":6}```

to 

**```{"data":[{"genre":"Music","count":6},{"genre":"Pop music","count":2},{"genre":"Rock music","count":1},{"genre":"Country music","count":1},{"genre":"Electronic music","count":1}],"totalLiked":6}```**

Achieved this by
- adding `YoutubeGenres.java` class. This is the object we send from backend to front end.
- adding `VideoGenreCount.java` class. This is an object of the form `{"genre":"Music","count":6}`

**Why reformat?**
Needs to be in this format to display d3.js bar chart.

### DIFF
--- a/server/src/main/java/com/google/musicanalysis/site/VideoGenreCount.java
+++ b/server/src/main/java/com/google/musicanalysis/site/VideoGenreCount.java
@@ -9,8 +9,4 @@ public class VideoGenreCount {
     this.genre = genre;
     this.count = count;
   }
-
-  public void incCount() {
-    this.count++;
-  }
 }
--- a/server/src/main/java/com/google/musicanalysis/site/VideoGenreCount.java
+++ b/server/src/main/java/com/google/musicanalysis/site/VideoGenreCount.java
@@ -1,0 +1,11 @@
+package com.google.musicanalysis.site;
+
+/** contains number of youtube videos under a music genre */
+public class VideoGenreCount {
+  private final String genre;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: server/src/main/java/com/google/musicanalysis/site/YoutubeGenres.java [hunk 1]
```diff
@@ -0,0 +1,14 @@
+package com.google.musicanalysis.site;
```
gracetian6, line 1: not sure if this is the correct area to put this java class
I added it here: `server/src/main/java/com/google/musicanalysis/site/YoutubeGenres.java`

server/src/main/java/com/google/musicanalysis/site/VideoGenreCount.java
```diff
@@ -0,0 +1,12 @@
+package com.google.musicanalysis.site;
+
+/** contains number of youtube videos under a music genre */
+public class VideoGenreCount {
+  public final String genre;
+  public int count;
```
gracetian6, line 6: also ... is it fine to have no `final/static` here for the `count` var? Otherwise can't increment the count.

gracetian6:
I'm going to make more changes!

PR #48, opened 2020-07-16
working MVP

seunomonije:
Working MVP code. 

https://drive.google.com/file/d/1vSdDrYJQ7nVTDfEEVq7X_bg6d97PV68Q/view?usp=sharing


client/js/analysis.js
```diff
@@ -1,25 +1,110 @@
...
+  removeAllChildNodes(list);
+  populateDonutCharts(json);
+  const time = populateComments(json);
+  determineOverall(json, time*1500);
+}
+
+/**
+ * Puts the charts on the screen
+ *
+ * @param {object} json the json object
```
laptou, line 38: what json object?

client/js/analysis.js
```diff
@@ -1,25 +1,110 @@
...
+      // Convert to string for type compatibility
+      addDonutChart(key, String(100 - map[key]*100));
+    }
+  }
+}
+
+/**
+ * Puts the comments on the screen
+ *
+ * @param {object} json the json object
```
laptou, line 53: what json object?

client/js/analysis.js
```diff
@@ -1,25 +1,110 @@
...
+ * Puts the charts on the screen
+ *
+ * @param {object} json the json object
+ */
+function populateDonutCharts(json) {
+  const map = json.perspectiveMap;
+  for (const key in map) {
+    if (key) { // this is here bc of the linter
+      // Convert to string for type compatibility
+      addDonutChart(key, String(100 - map[key]*100));
```
laptou, line 45: `String` constructor should be avoided; use `.toString()` instead

client/js/analysis.js
```diff
@@ -1,25 +1,110 @@
...
+
+/**
+ * Puts the charts on the screen
+ *
+ * @param {object} json the json object
+ */
+function populateDonutCharts(json) {
+  const map = json.perspectiveMap;
+  for (const key in map) {
+    if (key) { // this is here bc of the linter
```
laptou, line 43: that's not what the linter was trying to get you to do

the way that [`for...in`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) loops work is nuanced, so to ensure that it does what you are expecting it to do, you either need to check whether `key` is actually a property of `map` by checking if `map.hasOwnProperty(key)` is `true`:

```javascript
for (const key in map) {
  if (map.hasOwnProperty(key)) {
    addDonutChart(key, (100 - map[key] * 100).toString());
  }
}
```

or avoid using a `for...in` loop and use a `for...of` loop instead:

```javascript
for (const [key, value] of Object.entries(map)) {
  addDonutChart(key, (100 - value * 100).toString());
}
```

client/js/analysis.js
laptou: Also, why are you subtracting from 100? Are your percentages still reversed? Just use `calc()` in your CSS.

client/js/analysis.js
```diff
@@ -62,8 +147,11 @@ function addDonutChart(str, percent) {
   div.className = 'item donut';
   div.style.setProperty('--percent', percent);
 
+  // Convert back to actual percentage
+  const percentString = String(100 - Number.parseInt(percent));
```
laptou, line 151: `parseInt` is a global method, you don't need to do `Number.parseInt`

but also subtracting from 100 twice in the JS seems very janky compared to just subtracting from 100 in your CSS

laptou: and converting it to a string just to parse it back to a number also seems janky, why not just pass the number into your method?

client/js/analysis.js
seunomonije: calc messes up the keyframes

client/js/analysis.js
seunomonije: The IDE tells me `.setProperty(string, string)` is the correct way. One way or another there needs be conversion. This is so small of a thing it doesn't matter.

seunomonije:
Changes are resolved in MVP-extension pull request.

server/src/main/java/com/google/musicanalysis/types/AnalysisTrio.java
```diff
@@ -0,0 +1,19 @@
+package com.google.musicanalysis.types;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+public class AnalysisTrio {
```
geekster777, line 6: Just name this class `CommentAnalysis` instead to make it futureproof.

geekster777:
I've noticed you resolving a lot of comments without actually fixing them. In previous PR's as well. Please correct this behavior, as it's leading to a lot of extra work on my end to double check comments you've resolved.

geekster777:
Also just noticed that you don't have approval from your peers yet. Please do not add me and Jess as reviewers until they approve.

client/js/analysis.js
```diff
@@ -1,25 +1,113 @@
...
+};
+
+/**
+ * Fetches the data from the backend and populates the screen
+ *
+ */
+async function fetchResponse() {
+  const videoQuery = document.getElementById('searchbar').value;
+  const response = await fetch(`/api/analysis?name=${videoQuery}`);
+  const responseJson = await response.json();
```
gracetian6, line 17: 👍 oh didn't realize you could do `response.json()`

client/js/analysis.js
```diff
@@ -1,25 +1,113 @@
...
+  removeAllChildNodes(list);
+  populateDonutCharts(videoAnalysis.perspectiveMap);
+  // The +1 allows for determineOverall() to run correctly.
+  const time = renderComments(videoAnalysis.commentArray) + 1;
+  const overall = determineOverall(videoAnalysis.magnitudeAndScore.magnitude,
+      videoAnalysis.magnitudeAndScore.score);
+  setTimeout(() => {
+    addFeedbackResult(overall);
+  }, time*COMMENT_APPEARANCE_TIME);
+}
```
gracetian6, line 41: yess preloading! 😺

client/js/analysis.js
```diff
@@ -1,25 +1,113 @@
...
+ * which contains all of the data.
+ */
+function populationHandler(videoAnalysis) {
+  const charts = document.getElementById('charts');
+  const list = document.getElementById('list');
+
+  removeAllChildNodes(charts);
+  removeAllChildNodes(list);
+  populateDonutCharts(videoAnalysis.perspectiveMap);
+  // The +1 allows for determineOverall() to run correctly.
```
gracetian6, line 34: good comment

geekster777: Minimally helpful comment actually. Describe what the +1 does and how it interacts with `determineOverall`. Especially since `time` isn't passed to `determineOverall` anywhere.

It seems to be just used in the setTimeout below. I suggest restructuring this as:
```
const totalComments = renderComments(...);
const commentsRenderTime = totalComments * COMMENT_APPEARANCE_TIME;
...
setTimeout(() => {
  ...
}, commentsRenderTime + FEEDBACK_APPEARANCE_TIME);
```

client/js/analysis.js
```diff
@@ -42,7 +130,7 @@ function addFeedbackResult(result) {
   el.classList = 'center fade';
   const text = document.createElement('h1');
   text.setAttribute('id', 'overall');
-  text.innerHTML = 'Overall Response\n: ' + result;
+  text.innerHTML = 'Overall Response:\n ' + result;
```
geekster777, line 133: nit: add a `<br>` instead of a newline `\n`

client/js/analysis.js
```diff
@@ -58,12 +146,18 @@ function addFeedbackResult(result) {
 function addDonutChart(str, percent) {
   const div = document.createElement('div');
 
+  // Conversion math for CSS
+  percent = 100 - percent*100;
```
geekster777, line 150: Store this as its own const instead of modifying your parameter.
```
const percentRemaining = 100 - percent*100;
```

client/js/analysis.js
```diff
@@ -58,12 +146,18 @@ function addFeedbackResult(result) {
 function addDonutChart(str, percent) {
```
geekster777, line 146: The docstring above is no longer true. Percent is not a string. It is a decimal

geekster777: Also, name it "percentFull"

client/js/analysis.js
```diff
@@ -58,12 +146,18 @@ function addFeedbackResult(result) {
...
   div.className = 'item donut';
-  div.style.setProperty('--percent', percent);
+  div.style.setProperty('--percent', percent.toString());
+
+  // Convert back to actual percentage
+  const percentString = (100 - percent).toString();
 
   const header = document.createElement('h2');
-  header.innerText = str;
+  header.innerText = str + ': ' + percentString.substring(0, 2) + '%';
```
geekster777, line 160: nit: use format strings: \`${str}: ${percentString.substring(0, 2)}%\`

client/js/analysis.js
```diff
@@ -1,25 +1,113 @@
...
+  populationHandler(responseJson);
+}
+
+/**
+ * Handles populating all elements on the screen
+ *
+ * @param {object} videoAnalysis the object returned from the HTTP request
+ * which contains all of the data.
+ */
+function populationHandler(videoAnalysis) {
```
jessicaslaughter, line 27: nit: rename this renderingHandler (in general prefer "render" over "populate")

client/js/analysis.js
```diff
@@ -1,25 +1,113 @@
...
+    addFeedbackResult(overall);
+  }, time*COMMENT_APPEARANCE_TIME);
+}
+
+/**
+ * Puts the charts on the screen
+ *
+ * @param {Map<string, number>} map the map from the videoAnalysis object
+ */
+function populateDonutCharts(map) {
```
jessicaslaughter, line 48: nit: rename this renderDonutCharts

client/js/analysis.js
```diff
@@ -1,25 +1,113 @@
...
+  for (const key in map) {
+    if (Object.prototype.hasOwnProperty.call(map, key)) {
+      // Convert to string for type compatibility
```
jessicaslaughter, line 51: This comment isn't needed anymore

client/js/analysis.js
```diff
@@ -1,25 +1,113 @@
...
+      addDonutChart(key, map[key]);
+    }
+  }
+}
+
+/**
+ * Puts the comments on the screen
+ *
+ * @param {Array} array the comment array from the json response
+ * @returns {number} the stop index so that overall knows when to come in
```
jessicaslaughter, line 61: rephrase this comment to something like "the total number of comments rendered", this function doesn't depend on its caller so "overall" is out of context

client/js/analysis.js
```diff
@@ -1,25 +1,113 @@
...
-    const values = [
-      'TOXICITY',
-      'IDENTITY ATTACK',
-      'INSULT',
-      'PROFANITY',
-      'THREAT',
-      'FLIRTATION',
-    ];
-    addDonutChart(values[i], Math.floor(Math.random() * 61) + 30);
+  // so that the overall knows when to come in
```
jessicaslaughter, line 71: This comment isn't necessary

client/js/analysis.js
```diff
@@ -1,25 +1,113 @@
...
+}
+
+/**
+ * Determines the overall sentiment and adds it to the screen
+ *
+ * @param {number} magnitude the magnitude returned by Perspective API
+ * @param {number} score the score returned by Perspective API
+ * @returns {string} a string describing the magnitude and score in words
+ */
+function determineOverall(magnitude, score) {
```
jessicaslaughter, line 82: Consider renaming this to 'determineOverallSentiment' or just 'determineSentiment'. While reading through this PR, it felt confusing seeing "overall" referred to because it leaves the question "overall what?" unanswered.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -41,15 +43,16 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
-    String json = convertToJsonUsingGson(new AnalysisPair(perspectiveMap, commentsSentiment));
+    String json =
+        convertToJsonUsingGson(new AnalysisTrio(perspectiveMap, commentsSentiment, commentArray));
     res.setContentType("application/json;");
     res.getWriter().println(json);
   }
 
   /** @param arr the array that will be converted to json */
-  private String convertToJsonUsingGson(AnalysisPair pair) {
+  private String convertToJsonUsingGson(AnalysisTrio trio) {
```
jessicaslaughter, line 53: References to AnalysisTrio should be updated to CommentAnalysis

PR #51, opened 2020-07-17
added more to MVP

seunomonije:
Created a header for the page, added a video card that takes you to the video that the search matched, and fixed some styling stuff. 

Yall will see it during the demo later today.


client/js/analysis.js
```diff
@@ -19,38 +19,83 @@ async function fetchResponse() {
...
+
+  const card = document.createElement('div');
+  card.setAttribute('id', 'videocard');
+
+  const img = document.createElement('img');
+  img.classList = 'card-image';
+  img.setAttribute('src', `https://img.youtube.com/vi/${id}/sddefault.jpg`);
+  card.appendChild(img);
+
+  const videoInfo = document.createElement('div');
```
gracetian6, line 73: can you move this line 73 to right before line 84. That way I'm able to keep track of what's happening to `videoInfo` in one block

client/js/analysis.js
```diff
@@ -19,38 +19,83 @@ async function fetchResponse() {
...
+
+  const link = document.createElement('a');
+  link.setAttribute('href', `https://www.youtube.com/watch?v=${id}`);
+  link.setAttribute('target', '_blank');
+
+  const title = document.createElement('h3');
+  title.setAttribute('id', 'card-title');
+  title.innerText = name;
+
+  link.appendChild(title);
```
gracetian6, line 83: move this line to be with block above (after line 81) since `title` is being modified in that block?

client/js/analysis.js
```diff
@@ -19,38 +19,83 @@ async function fetchResponse() {
...
+ */
+function createCard(json) {
+  const id = json.videoId;
+  const name = json.videoInfo.name;
+  const channel = json.videoInfo.channel;
+
+  const el = document.getElementById('videocard-wrapper');
+
+  const card = document.createElement('div');
+  card.setAttribute('id', 'videocard');
```
gracetian6, line 66: the spacing / block chunks for this function is pretty good. I have two minor suggestions to make it better!

client/js/analysis.js
```diff
@@ -106,6 +152,16 @@ function removeAllChildNodes(parent) {
   }
 }
 
+/**
+ * Toggles the comment header
+ *
+ */
+function showCommentHeader() {
+  const el = document.getElementById('commentHeader');
+  el.classList.toggle('hidden');
```
geekster777, line 161: Provide a value for these toggles so you're not relying on html state being correct when you enter the function:
```
el.classList.toggle('hidden', false);
el.classList.toggle('fade', true);
```

server/src/main/java/com/google/musicanalysis/types/AnalysisGroup.java
```diff
@@ -3,17 +3,23 @@
 import java.util.ArrayList;
 import java.util.HashMap;
 
-public class AnalysisTrio {
+public class AnalysisGroup {
```
jessicaslaughter, line 6: This should remain "CommentAnalysis" or even "VideoAnalysis" since this class holds data and results obtained from analyzing a video. AnalysisGroup doesn't tell me much about what this class is for and makes it easy to start piling on fields that shouldn't really be here.

server/src/main/java/com/google/musicanalysis/types/NameAndChannel.java
```diff
@@ -0,0 +1,11 @@
+package com.google.musicanalysis.types;
+
+public class NameAndChannel {
```
jessicaslaughter, line 3: Consider renaming this "VideoInfo", since this class might not always hold just the name and channel of a video.

client/js/analysis.js
```diff
@@ -19,38 +19,82 @@ async function fetchResponse() {
 /**
  * Handles populating all elements on the screen
  *
- * @param {object} json the json object
+ * @param {object} json the json object from backend
```
geekster777, line 22: You need a better variable name here

client/js/analysis.js
```diff
@@ -92,6 +136,7 @@ function determineOverall(json, time) {
 
   setTimeout(() => {
     addFeedbackResult(isClear + tone);
+    createCard(json);
```
geekster777, line 139: Pass the minimal parts of this json object instead of passing the whole thing.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -120,6 +128,24 @@ private NLPResult analyzeWithNLP(String text) throws IOException {
     return perspectiveResults;
   }
 
+  private VideoInfo getVideoInfo(String response) {
+
+    // Accessing the items JSON Array
+    JsonElement jElement = JsonParser.parseString(response);
+    JsonObject jObject = jElement.getAsJsonObject();
+    JsonArray itemsArray = jObject.getAsJsonArray("items");
+    JsonElement el = itemsArray.get(0);
```
geekster777, line 137: What is el? Give this better variable name, like "firstVideo"

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -120,6 +128,24 @@ private NLPResult analyzeWithNLP(String text) throws IOException {
...
+
+    // Grabbing the name and channel
+    JsonObject object = el.getAsJsonObject();
+    JsonObject data = object.getAsJsonObject("snippet");
```
geekster777, line 141: same

PR #53, opened 2020-07-20
Can now enter just id

seunomonije:
Made it so that you can enter a video id or the video name. Will help with the quota as well as if users want a specific video.


server/src/main/java/com/google/musicanalysis/api/youtube/YoutubeRequest.java
```diff
@@ -38,6 +39,12 @@ public String getResult() throws MalformedURLException, IOException {
 
     URL url = buildUrl();
 
+    HttpURLConnection con = (HttpURLConnection) url.openConnection();
+    int response = con.getResponseCode();
+    if (response > 400) {
```
laptou, line 44: check if response != 200 instead, a 301 response won't be very useful to you either

server/src/main/java/com/google/musicanalysis/api/youtube/YoutubeRequest.java
```diff
@@ -38,6 +39,12 @@ public String getResult() throws MalformedURLException, IOException {
...
+      return "unreadable";
```
laptou, line 45: throw an exception instead of returning this string

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -21,23 +21,32 @@
   protected void doGet(HttpServletRequest req, HttpServletResponse res)
       throws ServletException, IOException {
 
-    String videoName = req.getParameter("name");
+    String input = req.getParameter("name");
 
-    assert videoName == null : "Something went wrong with sending to backend.";
+    assert input == null : "Something went wrong with sending to backend.";
```
gracetian6, line 26: throw errors instead? (Though I guess this is a type of error...)

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -21,23 +21,32 @@
...
+    } catch (IOException err) {
+      // We need to search for the id and get the comments again
+      videoArgs.put("q", input);
+      videoArgs.put("type", "video");
+      String videoIdJson = new YoutubeRequest("search", videoArgs).getResult();
+      videoId = getVideoId(videoIdJson);
+
+      commentArgs.replace("videoId", videoId);
+      commentsJson = new YoutubeRequest("commentThreads", commentArgs).getResult();
+    }
```
gracetian6, line 49: Nick will prob say something about how doing try/catch is lazy/bad practice. Could you catch the error by adding an if statement to check whether a variable is null?

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -210,6 +219,10 @@ private String convertToString(ArrayList<String> comments) {
 
     for (String comment : comments) {
       comment = comment.replace("\"", "");
+      
+      if (comment.length() - 1 <= 0){
```
gracetian6, line 223: `comment.length() <= 1` is more readable???

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
laptou: `try`/`catch` is not bad practice. Why would it be?

Also, he's catching an `IOException`. That has nothing to do with whether any variable is `null`.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
laptou: `assert` statements are almost the same as throwing an error. The main difference is that `assert` statements can be ignored when running in production, which means that `input` will not be checked for `null` when you deploy. All of that can be configured, but watch out.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
geekster777: You should also leave a comment why you're skipping 1 and 0 letter long comments

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
geekster777: +1. `try/catch` is only bad when you're using it as a cure-all to catch errors in your code (null pointer exceptions, divide by zero errors, etc). It's necessary when interacting with a function that explicitly throws errors, such as `IOException`'s. It's generally best practice to minimize the surface area of your `try/catch`, which is being done here.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -21,23 +21,32 @@
...
-    videoArgs.put("q", videoName);
-    videoArgs.put("type", "video");
-    String videoIdJson = new YoutubeRequest("search", videoArgs).getResult();
-    String videoId = getVideoId(videoIdJson);
+    String videoId = input; // assume user enters id
 
     commentArgs.put("part", "snippet");
-    commentArgs.put("videoId", videoId);
-    String commentsJson = new YoutubeRequest("commentThreads", commentArgs).getResult();
+    commentArgs.put("videoId", input);
```
geekster777, line 36: Are video id's a standard length/format? You can likely skip most "search strings" just by checking if the video id is 11 characters with no spaces.

seunomonije: I'm not sure I'm following...what do you mean by skip most search strings?
I also though of checking the id w/no spaces, but then some channels like "deestroying", "RomanAtwood", or any 11 word video would match.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
seunomonije: I'll just check on the front-end for more clarity.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -33,7 +31,7 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
     String videoId = input; // assume user enters id
 
     commentArgs.put("part", "snippet");
-    commentArgs.put("videoId", input);
+    commentArgs.put("videoId", input);
```
jessicaslaughter, line 34: nit: remove the extra space after ;

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -219,8 +217,9 @@ private String convertToString(ArrayList<String> comments) {
 
     for (String comment : comments) {
       comment = comment.replace("\"", "");
-      
-      if (comment.length() - 1 <= 0){
+      // comment.replace() could potentially give an empty string,
```
jessicaslaughter, line 220: This comment just tells me how replace() and substring() work, and doesn't tell me why a comment with a length of 0 or 1 shouldn't be modified below (that comments with 0 or 1 characters aren't considered sentences)

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
geekster777: Yeah, but looking up a video with id="RomanAtwood" should return a 404, and then you can fall back on the default searching behavior. As it stands, if I send you a video name, you're going to make 2 requests to youtube anyways (one as if I sent an id and another as if I sent the video name), so this would be be a net positive in saving you quota.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
geekster777: Please add a comment. Something like "One letter strings can't be properly concatenated with the other comments because [reasons]". Right now I'm not sure why you're skipping one letter comments.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -217,8 +217,10 @@ private String convertToString(ArrayList<String> comments) {
 
     for (String comment : comments) {
       comment = comment.replace("\"", "");
-      // comment.replace() could potentially give an empty string,
-      // which messes up lastCharacter.
+      // This does two things:
```
jessicaslaughter, line 220: This still only tells me what (1) substring() does when given a 0 value, and (2) how 'continue' works. I still don't know why one letter comments are being skipped, since you could've just treated one letter comments specially and skipped the substring() call. This comment should say something like "One letter comments aren't concatenated with the larger string because we don't consider them to be sentences" or "One letter comments can't be concatenated with the other comments because they hurt the overall analysis on the larger string". Just throwing ideas out because I'm not sure what the reason is

seunomonije: Ah bet. I see what you're saying now.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
seunomonije: Ah makes sense

seunomonije: Can you look at what I did to see if it's what you were getting at?

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -14,30 +14,51 @@
...
+    String videoId = input; // assume user enters id
 
     commentArgs.put("part", "snippet");
-    commentArgs.put("videoId", videoId);
-    String commentsJson = new YoutubeRequest("commentThreads", commentArgs).getResult();
+    commentArgs.put("videoId", input);
+    String commentsJson;
+    
+    // if the input doesn't look like a standard id
+    if (thereIsWhiteSpace(input) || input.length() > 11) {
```
geekster777, line 40: nit: check if the length is `!= 11` since a 10 character input is also going to be a search.

geekster777, line 40: Invert this logic. You want to check first if you can get a video by this "id". If it comes back with a 404, then you want to use the search API.

geekster777: You should also separate this logic out into helper functions now that this function is getting long. Something like:
```
Comments comments;
if (videoParam.length() == 11 && !thereIsWhitespace(videoParam)) {
  try {
    comments = getCommentsFromVideoId(videoParam);
  } catch {
    String videoId = getFirstVideoFromSearch(videoParam);
    comments = getCommentsFromVideoId(videoId);
  }
}
```

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -210,6 +231,14 @@ private String convertToString(ArrayList<String> comments) {
 
     for (String comment : comments) {
       comment = comment.replace("\"", "");
+      // This does makes sure of two things:
+      // 1. commment.substring(comment.length() - 1)
+      //    doesn't throw an error, which ruins the loop
+      // 2. One letter comments should not be considered as sentences since
+      //    they bring no value to the analysis.
```
geekster777, line 238: I'd just stick with this comment. The one before is nice, but it's implied that empty values should be ignored normally,.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -14,30 +15,42 @@
...
+    String commentsJson;
+    
+    // Test if its a youtube id from the beginning
+    if (userInput.length() == 11 && !thereIsWhiteSpace(userInput)) {
+        try {
+            commentsJson = new YoutubeRequest("commentThreads", commentArgs).getResult();
+        } catch (IOException err) {
+            videoId = getFirstVideoFromSearch(userInput);
+            commentArgs.replace("videoId", videoId);
+            commentsJson = new YoutubeRequest("commentThreads", commentArgs).getResult();
```
geekster777, line 47: nit: this code is identical to what's in your `else` clause. You can just use an if statement where you check if commentsJson is null after trying to get comments with the video id, and only leave a comment in your catch
```
if (/** is a video id */) {
  try {
    commentsJson = getVideoComments(videoId);
  } catch {
    // It wasn't a video id.
  }
}

if (!commentsJson) {
  // Search video and get comments.
}
```

PR #54, opened 2020-07-20
initial caching attempt

seunomonije:
First attempt at creating a cache. Thank you MDN for making this easy on me lol. 

#### Known Issues:
- if the user enters a video id, then a video name that matches with that id, the cache will fetch new data rather than take from itself. I would like to fix this to have as little API calls as possible. 
- I would like to add something that gives more control over when the cache fetches new data, for example, if the video was recent the sentiments might change and therefore will need fresher data, but if the video is older then you can use older data since the sentiment is already established.


client/js/analysis.js
```diff
@@ -10,10 +10,13 @@ window.onload = function() {
...
-  const response = await fetch(`/api/analysis?name=${param}`);
-  const value = await response.json();
-  console.log(value);
-  populationHandler(value);
+  try {
+    // see caching.js for this function
+    const response = await getData(`/api/analysis?name=${param}`);
+    populationHandler(response);
+  } catch (error) {
+    console.log(error);
```
laptou, line 18: nit: use `console.error` for errors

client/js/caching.js
```diff
@@ -0,0 +1,63 @@
...
+  const cacheVersion = 1;
+  const cacheName = `analysisCache-${cacheVersion}`;
+  let cachedData = await getCachedData(cacheName, url);
+
+  if (cachedData) {
+    console.log('Retrieved cached data');
+    return cachedData;
+  }
+
+  console.log('Fetching fresh data');
```
gracetian6, line 18: delete these console.log status messages

client/js/caching.js
```diff
@@ -0,0 +1,63 @@
...
+ * Deletes old caches so that the cache doesn't
+ * take up too much space
+ *
+ * @param {Cache} currentCache the cache to traverse through
+ */
+async function deleteOldCaches(currentCache) {
+  const keys = await caches.keys();
+
+  for (const key of keys) {
+    const correctCache = 'analysisCache-' === key.substr(0, 14);
```
gracetian6, line 56: nit: parentheses?
```
    const correctCache = ('analysisCache-' === key.substr(0, 14));
```

client/js/caching.js
```diff
@@ -0,0 +1,63 @@
...
+ * @param {string} url the url string to send
+ * @returns {object} the response object
+ */
+async function getData(url) {
+  const cacheVersion = 1;
+  const cacheName = `analysisCache-${cacheVersion}`;
+  let cachedData = await getCachedData(cacheName, url);
+
+  if (cachedData) {
+    console.log('Retrieved cached data');
```
geekster777, line 14: remove debug statement.

client/js/caching.js
```diff
@@ -0,0 +1,63 @@
...
+ * @param {Cache} cacheName the name of the cache to search
+ * @param {string} url the url string, which is the key
+ * @returns {object} response object from the cache
+ */
+async function getCachedData(cacheName, url) {
+  const cacheStorage = await caches.open(cacheName);
+  const cachedResponse = await cacheStorage.match(url);
+
+  if (!cachedResponse || !cachedResponse.ok) {
+    return false;
```
geekster777, line 40: return `null` instead of false.

client/js/caching.js
geekster777, line 56: use `key.startsWith(...)` instead of making a substring.

client/js/analysis.js
```diff
@@ -10,10 +10,12 @@ window.onload = function() {
  */
 async function fetchResponse() {
   const param = document.getElementById('searchbar').value;
-  const response = await fetch(`/api/analysis?name=${param}`);
-  const value = await response.json();
-  console.log(value);
-  populationHandler(value);
+  try {
+    const response = await getData(`/api/analysis?name=${param}`);
+    populationHandler(response);
```
geekster777, line 15: nit: you shouldn't have this function in your try/catch. Try to make your try/catch as narrowly scoped as possible, otherwise this will just quietly swallow any errors from populationHandler. Better to do:
```
let response;
try {
  response = ...;
} catch (e) {
  ...
}
if (response) {
  populationHandler(response);
}
```

PR #57, opened 2020-07-22
Cache encryption

seunomonije:
Working file encryption for the cache. Included an example use case AnalysisServlet.java that will be filled with a request url in the next pull request


server/src/main/java/com/google/musicanalysis/cache/AnalysisCache.java
```diff
@@ -1,87 +1,112 @@
...
+    // map.get() returns null if not match
+    return this.responseMap.get(requestUrl);
+  }
+
+  public void delete(String requestUrl) {
+    this.responseMap.remove(requestUrl);
+  }
+
+  private int createFile() {
+    try { // logic here is going to improve which is why it's int
```
laptou, line 93: Use an `enum` or exceptions to represent complex results in Java

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -58,6 +64,21 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
+          new AnalysisGroup(perspectiveMap, commentsSentiment, commentArray, videoId, videoInfo));
+      cache.close();
+    } catch (InvalidKeyException
+        | IllegalBlockSizeException
+        | BadPaddingException
+        | NoSuchAlgorithmException
+        | NoSuchPaddingException e) {
+      e.printStackTrace();
+    }
+
```
laptou, line 81: This code definitely works, but it seems like it would become increasingly slow if you do an I/O operation that rewrites the whole file on every request. What are your plans here?

seunomonije: I had two thoughts:
  1. Rather than rewrite the file, I can just append the latest entry and delete any existing, stale matches.
  2. Store all the cached data that the user gets in that session, and then when the session ends rewrite the entire file.
  3. Do both.

server/src/main/java/com/google/musicanalysis/cache/AnalysisCache.java
```diff
@@ -1,87 +1,115 @@
...
-    public AnalysisGroup search(String requestUrl) {
-        // map.get() returns null if not match
-        return this.responseMap.get(requestUrl);
+  public void close() throws InvalidKeyException, IllegalBlockSizeException {
+    try {
+      cipher.init(Cipher.ENCRYPT_MODE, generateKey());
+      SealedObject sealedObject = new SealedObject(this.responseMap, cipher);
+      CipherOutputStream outFile =
+          new CipherOutputStream(
+              new BufferedOutputStream(new FileOutputStream("cachedData.txt")), cipher);
```
geekster777, line 72: Put this filename in a const at the top of the file

server/src/main/java/com/google/musicanalysis/cache/AnalysisCache.java
```diff
@@ -1,87 +1,115 @@
...
-        } catch (IOException e) {
-            e.printStackTrace();
-            return;
-        } catch (ClassNotFoundException c) {
-            System.out.println("Analysis Group class not found");
-            c.printStackTrace();
-            return;
-        }
-    }
+  public void open() throws InvalidKeyException, IllegalBlockSizeException, BadPaddingException {
```
geekster777, line 41: Instead of needing to open this file every time someone makes a web request, can we keep that cache persistent between calls? Maybe only periodically write to the file? File reads/writes are very expensive compared to reads/writes from memory.

server/src/main/java/com/google/musicanalysis/cache/AnalysisCache.java
```diff
@@ -1,87 +1,115 @@
...
+    FileStatus file = createFile(); // this will have use later.
 
-    public void add(String requestUrl, AnalysisGroup responseData) {
-        this.responseMap.put(requestUrl, responseData);
-    }
+    try {
+      cipher.init(Cipher.DECRYPT_MODE, generateKey());
+      CipherInputStream inFile =
+          new CipherInputStream(
+              new BufferedInputStream(new FileInputStream("cachedData.txt")), cipher);
```
geekster777, line 50: Keep this filename in a const.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -58,6 +64,21 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
     HashMap<String, String> perspectiveMap = analyzeWithPerspective(cumulativeComments);
     NLPResult commentsSentiment = analyzeWithNLP(cumulativeComments);
 
+    try {
+      AnalysisCache cache = new AnalysisCache();
+      cache.open();
+      cache.add(
+          "example key",
```
geekster777, line 71: Use the video id as a key here.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
geekster777: Consider making your cache persist in memory between web requests. I'm a bit hazy with how AppEngine handles lifecycles for your app, but I think this would be possible and keep your user from waiting on file reads and periodically flushing your cache to the filesystem.

PR #60, opened 2020-07-23
Vanilla bars

gracetian6:
- added svg bars from hard coded data
- added two blocks side by side within vis-container. 
  - Left html block `bar-text` will contain labels for each bar. 
  - right svg block contained within `graph` contains bars.

image

image


client/youtube-genre.html
```diff
@@ -4,19 +4,29 @@
...
-          <input type="number" name="numVideos" id="numVideos" value="10" placeholder="Number of Liked Videos"/>
-          <input type="submit" onclick="displayMusicGenre()"/>
-        </p>
-      </div>
+    <div class="title hero" id="banner"> 
+      <h1>What music do you listen to on YouTube? </h1>
+      <p>
+        Choose number of liked videos to retrieve:
+        <input type="number" name="numVideos" id="numVideos" value="10" placeholder="Number of Liked Videos"/>
+        <input type="submit" onclick="displayMusicGenre()" class="btn"/>
```
laptou, line 16: nit: prefer `<button type="submit">` to `<input type="submit">`

client/js/bar.js
```diff
@@ -0,0 +1,35 @@
...
+const maxCount = 3;
+
+const GRAPH_HEIGHT = 100;
+const BAR_CONTAINER_HEIGHT = GRAPH_HEIGHT / counts.length;
+
+// bar container contains a bar. It is composed of bar padding and fill
+const BAR_PERCENT_FILL = 0.8;
+const BAR_PERCENT_PADDING = 1 - BAR_PERCENT_FILL;
+const BAR_THICKNESS = BAR_PERCENT_FILL * BAR_CONTAINER_HEIGHT;
+const BAR_UNIT_LENGTH = 100 / maxCount;
```
seunomonije, line 18: Know we talked about this already, but this actually looks very clean and is a good way to set these values.

client/js/bar.js
```diff
@@ -0,0 +1,35 @@
+/* constructs bar graph from scratch */
+
+const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+svg.setAttribute('viewBox', '0 0 100 100'); // change height in css
+document.getElementById('graph').appendChild(svg);
+
+// hard coded data for bar chart for now
+const counts = [1, 3, 1, 2];
```
geekster777, line 8: nit: give this a more descriptive name, like chartValues

client/js/bar.js
```diff
@@ -0,0 +1,35 @@
...
+const maxCount = 3;
```
geekster777, line 9: You should compute this with `Math.max(...counts);`

geekster777, line 9: Use SCREAMING_SNAKE_CASE

client/js/bar.js
```diff
@@ -0,0 +1,35 @@
+/* constructs bar graph from scratch */
+
+const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+svg.setAttribute('viewBox', '0 0 100 100'); // change height in css
+document.getElementById('graph').appendChild(svg);
```
geekster777, line 5: This should go below the file consts below.

client/js/bar.js
```diff
@@ -0,0 +1,35 @@
...
+  svg.appendChild(g);
+  g.setAttribute('class', 'bar');
+
+  const bar = document.createElementNS('http://www.w3.org/2000/svg', 'rect');
+  g.appendChild(bar);
+  bar.setAttribute('width', BAR_UNIT_LENGTH * counts[i]);
+  bar.setAttribute('height', BAR_THICKNESS);
+  bar.setAttribute('x', GRAPH_LEFT_PADDING);
+  bar.setAttribute('y', GRAPH_TOP_PADDING + BAR_CONTAINER_HEIGHT * i);
+}
```
geekster777, line 35: This should be a function that you call at the end of the file:
```
function createChart(chartValues) {
   ...
}
createChart(DEFAULT_CHART_VALUES);
```

client/js/bar.js
gracetian6: new changes implement this. Is it good to put it after file consts because all file consts pertain to the svg set up?

geekster777: Yes. Consider file const's to be parameters to the file that a developer will want to toggle without reading anything else. The only thing that should go above file const's are imports.

client/js/bar.js
```diff
@@ -0,0 +1,44 @@
+/* constructs bar graph from scratch */
+
+const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+svg.setAttribute('viewBox', '0 0 100 100');
```
geekster777, line 4: This should be in your function. Not created out here.

client/js/bar.js
```diff
@@ -0,0 +1,44 @@
...
+const GRAPH_TOP_PADDING = (BAR_PERCENT_PADDING / 2) * BAR_CONTAINER_HEIGHT;
+const GRAPH_LEFT_PADDING = 2.5;
+
+document.getElementById('graph').appendChild(svg);
+
+/**
+ * creates SVG bar chart given chart values
+ * @param {int[]} CHART_VALUES
+ */
+function createBarChart(CHART_VALUES) {
```
geekster777, line 29: parameters should be snakeCase. Also, you should move any file level const that is derived from CHART_VALUES into this function (with snakeCase) and derive them from this parameter instead.

client/js/bar.js
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * creates SVG bar chart given chart values
+ * @param {int[]} chartValues
+ */
+function createBarChart(chartValues) {
+  const maxChartValues = Math.max(...chartValues);
+  const barContainerHeight = GRAPH_HEIGHT / chartValues.length;
+
+  const barThickness = BAR_PERCENT_FILL * barContainerHeight;
+  const BAR_UNIT_LENGTH = 100 / maxChartValues;
```
geekster777, line 23: this should be snakeCase

client/js/bar.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // top graph padding depends on bar padding
+  const GRAPH_TOP_PADDING = (BAR_PERCENT_PADDING / 2) * barContainerHeight;
```
geekster777, line 26: same

client/js/bar.js
```diff
@@ -0,0 +1,47 @@
...
+  const GRAPH_LEFT_PADDING = 2.5;
```
geekster777, line 27: Move this to the top of the file.

PR #61, opened 2020-07-23
Vanilla bars with category text

gracetian6:
**Added left labels for each corresponding right bar.** Data is still hardcoded in JS.

Achieved this by 
- creating multiple div block containers for each label. This text container should align with the corresponding bar.
- vertically centering text within its container

image

image

### DIFF
--- a/client/css/bar.css
+++ b/client/css/bar.css
@@ -44,7 +44,6 @@
   display: flex;
   flex-direction: column;
   justify-content: center;
-  float: right;
 
   text-align: right;
 }
--- a/client/css/bar.css
+++ b/client/css/bar.css
@@ -25,7 +25,7 @@
 
   transition: fill .3s ease;
 
-  fill: var(--color-accent-primary-2); /* how to make global? */
+  fill: var(--color-accent-primary-2);
 }
 
 .bar:hover, .bar:focus {
@@ -33,8 +33,6 @@
 }
 
 #bar-text {
-  position: relative;
-
   display: inline-block;
   width: 25%;
 
--- a/client/css/bar.css
+++ b/client/css/bar.css
@@ -6,6 +6,11 @@
   height: 100vh;
 }
 
+.svg {
+  height: 100%;
+  width: 100%;
+}
+
 #vis {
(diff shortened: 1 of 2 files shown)


client/css/bar.css
```diff
@@ -35,4 +35,16 @@
...
+  vertical-align: top;
+}
+
+.category {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  float: right;
+
+  text-align: right;
```
laptou, line 49: Why do you have `float: right`?

seunomonije: ^

client/js/bar.js
```diff
@@ -4,13 +4,27 @@ const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
...
 document.getElementById('graph').appendChild(svg);
 
+// textBlock should align vertically with bar chart
+const barTextContainer = document.getElementById('bar-text');
+barTextContainer.setAttribute('height', svg.getBoundingClientRect().height);
+
 // hard coded data for bar chart for now
 const counts = [1, 3, 1, 2];
 const maxCount = 3;
+const genres = [
```
geekster777, line 14: nit: this should be in SCREAMING_SNAKE_CASE

client/js/bar.js
```diff
@@ -4,13 +4,27 @@ const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
 svg.setAttribute('viewBox', '0 0 100 100'); // change height in css
 document.getElementById('graph').appendChild(svg);
 
+// textBlock should align vertically with bar chart
+const barTextContainer = document.getElementById('bar-text');
+barTextContainer.setAttribute('height', svg.getBoundingClientRect().height);
```
geekster777, line 9: This should go after the consts.

geekster777, line 9: This is a bit fragile. You should be able to solve this with css. Try something like `height: 100%` and let me know if you need help debugging.

gracetian6: TODO on tuesday

PR #62, opened 2020-07-24
Create basic table for chart

laptou:
- There is now a table that shows the top 15 songs in order based on where the user hovers in the chart

https://drive.google.com/file/d/1-qtus46UcrB6PjAmDhj87Y3IPR45nBQA/view?usp=sharing


client/js/gdpr/table.js
```diff
@@ -0,0 +1,80 @@
...
+
+      if (rank === undefined || rank === null) {
+        continue;
+      }
+
+      ranking.push({key, rank, selected});
+    }
+
+    // sort by rank ascending
+    ranking.sort((a, b) => a.rank < b.rank ? -1 : 1);
```
seunomonije, line 49: Is `rank` stored as a number? If so, why not `ranking.sort((a,b) => a.rank - b.rank)`?

laptou: That would work too, it just didn't occur to me. It seems to have the same behaviour, but your version might be easier to read.

PR #64, opened 2020-07-24
Use custom element for chart

laptou:
- Use a [custom element](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) for the GDPR chart instead of encapsulating all of the chart's data in closures
- [New chart](https://github.com/googleinterns/step99-2020/pull/64/files#diff-e5d902d30fa9b9c35faa1c36e9bd5fd0R28) is functionally the same, but code is tidier


client/js/gdpr/chart.js
```diff
@@ -4,73 +4,80 @@ const RUN_SCALE_X = 30;
...
+      this.rows.push([]);
     }
 
-    seriesContainer.append(series);
-    seriesElements.push(series);
-    index++;
-  }
+    for (const history of histories.values()) {
+      for (let row = 0; row < dates.length; row++) {
+        this.rows[row + 1].push(history[row] || null);
```
seunomonije, line 61: At first I was wondering why you left the first row empty in `this.rows`, but now I see that you filled it on line 53.

client/js/gdpr/chart.js
```diff
@@ -79,289 +86,322 @@ export function createChart(container, histories, dates) {
...
+
+    const seriesContainer = document.createElementNS(SVG_NS, 'g');
+
+    /** @type {SVGGElement[]} */
+    const seriesElements = [];
+
+    let index = 0;
+
+    for (const history of this.histories.values()) {
+      const hue = index * 15 % 360;
```
seunomonije, line 169: I like how you broke everything up here

laptou:
> If this PR creates bugs that PR #65 fixes, wouldn't it be better to just get rid of this PR and merge the two together into one PR that works?

@gracetian6 I guess; I was just trying to split everything into small pieces. However, I also don't really consider this PR to introduce a bug, it's just not intended to be complete by itself.

client/js/gdpr/chart.js
```diff
@@ -79,289 +90,323 @@ export function createChart(container, histories, dates) {
...
-  svg.addEventListener('mouseleave', () => clearHover());
-  scrollContainer.addEventListener('scroll', () => clearHover());
+  createChart() {
+    const svg = document.createElementNS(SVG_NS, 'svg');
+    svg.setAttribute('class', 'chart');
+    svg.setAttribute(
+        'viewBox',
+        // shift down so that lines are vertically centered
+        `0 ${RUN_SCALE_Y * 0.5} ` +
+        `${this.dates.length * RUN_SCALE_X} ${(NUM_POSITIONS + 0.5) * RUN_SCALE_Y}`,
```
gracetian6, line 161: lint says to make this line less chars

client/js/gdpr/chart.js
```diff
@@ -4,73 +4,84 @@ const RUN_SCALE_X = 30;
...
+      this.rows.push([]);
     }
 
-    seriesContainer.append(series);
-    seriesElements.push(series);
-    index++;
-  }
+    for (const history of histories.values()) {
+      for (let row = 0; row < dates.length; row++) {
+        this.rows[row + 1].push(history[row] || null);
```
geekster777, line 65: Why `+1`? Are rows 1 indexed? If so, please specify this with a comment.

client/js/gdpr/chart.js
```diff
@@ -79,289 +90,323 @@ export function createChart(container, histories, dates) {
...
+
+    const seriesContainer = document.createElementNS(SVG_NS, 'g');
+
+    /** @type {SVGGElement[]} */
+    const seriesElements = [];
+
+    let index = 0;
+
+    for (const history of this.histories.values()) {
+      const hue = index * 15 % 360;
```
geekster777, line 174: I think this was addressed in an earlier PR, but add a comment explaining that you want hues to be 15 degrees apart on the color wheel so we have context as to what 15 and 360 here stand for.

laptou: argh why does it keep disappearing

client/js/gdpr/chart.js
laptou: I commented on line 56, `// first row is a header row containing keys`

laptou: I commented on line 56, `// first row is a header row containing keys`

client/js/gdpr/chart.js
laptou: fixed

PR #65, opened 2020-07-24
Reconnect chart to table

laptou:
- #64 broke the connection between the chart and the table; this PR fixes it
- Use better identifiers for CSS containers in chart


client/js/gdpr/chart.js
```diff
@@ -21,8 +21,11 @@ export class GdprChart extends HTMLElement {
 
     this.hoverState = {seriesIndex: null, x: null, y: null};
 
+    this.mainContainer = document.createElement('div');
+    this.mainContainer.id = 'main-container';
+
     this.scrollContainer = document.createElement('div');
-    this.scrollContainer.classList.add('chart-scroll-container');
+    this.scrollContainer.id = 'scroll-container';
```
gracetian6, line 28: is there a reason why changing the class `chart-scroll-container` into the id `scroll-container` is a better design choice?

client/js/gdpr/chart.js
```diff
@@ -123,6 +127,7 @@ export class GdprChart extends HTMLElement {
             key,
           },
           bubbles: true,
+          composed: true,
```
gracetian6, line 130: is it possible to add a comment somewhere about why adding `composed` is important?

client/js/gdpr/chart.js
laptou: Not really; I accidentally ended up implementing the same functionality twice b/c I was switching branches so much. Both styles are functionally equivalent, but I figured I might as well use an ID since this element is unique within the shadow root.

client/js/gdpr/chart.js
laptou: I could do that, but I think that since `composed` essentially only has one function (to allow events to cross the shadow root boundary) it should be up to the developer to understand how custom events work.

PR #66, opened 2020-07-24
Convert table to custom element

laptou:
- Converts the table to a custom element

Using custom elements provides a more convenient API for adding the table and chart to the document: they are already in the document, and all I need to do is load the data into them. I also think that it provides better encapsulation because the styles for each component are scoped to the custom element and events don't bubble by default.


laptou:
> is there a reason why making it a custom element is useful? Please add to your PR description!

@gracetian6 Done

PR #67, opened 2020-07-24
Run lint all the time

laptou:
- Linter now runs on all branches / PRs, not just the ones based on `master`
  - This will allow lint errors / warnings to be caught sooner


laptou:
> Do we need to worry about "Unchanged files with check annotations" ?

Only if you are the one who created those files; it's essentially a reminder that those files still have issues but you did not edit them in that PR.

PR #68, opened 2020-07-27
Cosmetic changes to table

laptou:
- Very minor cosmetic changes to table

Screenshot 2020-07-27 at 15 29 09 - Display 1


client/gdrive-picker-helloworld.html
```diff
@@ -0,0 +1,105 @@
...
+limitations under the License.
+-->
+<!-- [START picker_hello_world] -->
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml">
+  <head>
+    <meta charset="utf-8" />
+    <title>Google Picker Example</title>
+
+    <script type="text/javascript">
```
seunomonije, line 23: JS in html file rather than it's own? I respec it, but not sure what others will say

client/gdrive-picker-helloworld.html
```diff
@@ -0,0 +1,105 @@
...
+
+    // The Browser API key obtained from the Google API Console.
+    // Replace with your own Browser API key, or your own key.
+    var developerKey = 'xxxxxxxYYYYYYYY-12345678';
```
seunomonije, line 27: Was thinking of doing this for my own website, but allowing users to add their own API keys on the front end then using them in the code would be pretty cool. Just a thought.

client/gdrive-picker-helloworld.html
gracetian6: Yea I am also confused about this.

gracetian6:
okay I see that this hello world file is deleted in the next PR. I don't think it was necessary to send this out for review and this and the other file could have been merged or at least deleted in this PR

client/gdrive-picker-helloworld.html
laptou: Ah, I didn't mean to commit that file. That is just a sample file that I copied from somewhere else.

PR #71, opened 2020-07-27
Comment filtering

seunomonije:
For some reason I forgot to push these commits to other prs, I'll get that fixed.

Otherwise, I buffed the comment filtering and added likes for each retrieved comment which I can use for weight analysis later


server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -231,4 +243,24 @@ private String convertToString(ArrayList<String> comments) {
...
+  private String filterComment(String comment) {
+    String filteredComment = comment;
+
+    // Handles non-ASCII characters, newlines,
+    // trimmed quotes, embedded escaped quotes.
+    // Order matters.
+    filteredComment = filteredComment.replaceAll("[^\\x00-\\x7F]", "");
+    filteredComment = String.join("\n", filteredComment.split("\\\\n"));
+    filteredComment = filteredComment.substring(1, filteredComment.length() - 1);
+    filteredComment = filteredComment.replace("\\\"", "");
```
laptou, line 262: Is the analysis supposed to work on non-English text? Because filtering out all non-ASCII characters will create a lot of problems with that.

server/src/main/java/com/google/musicanalysis/types/CommentLikes.java
```diff
@@ -0,0 +1,14 @@
+package com.google.musicanalysis.types;
+
+import java.io.Serializable;
+
+/** Groups together a comment with its likes */
+public class CommentLikes implements Serializable {
```
gracetian6, line 6: nice name

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -231,4 +243,24 @@ private String convertToString(ArrayList<String> comments) {
...
     return res.toString();
   }
+
+  /**
+   * Filters out a comment string so that it's readable and doesn't break the frontend.
+   *
+   * @param comment The comment to be filtered.
+   * @return A properly formatted comment
+   */
+  private String filterComment(String comment) {
```
gracetian6, line 253: could this be a method within CommentLikes.java class instead?

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
seunomonije: Good call. Will take out non-ASCII and replace with just emojis

server/src/main/java/com/google/musicanalysis/types/CommentLikes.java
jessicaslaughter: I would consider calling this "Comment" instead of "CommentLikes" because each comment "has" likes and "has" text content. You're already using it throughout your files with an associated variable name of "comments" or "commentsArray" as well

gracetian6:
I need this github PR to be merged into master so that things will run when I pull from master 😿

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
seunomonije: I like the idea, but the Comment.java class is supposed to represent a data "type". I would prefer to have no methods inside it.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -199,8 +200,17 @@ private String getVideoId(String response) {
...
+      Integer likes =
+          el.getAsJsonObject()
+              .getAsJsonObject("snippet")
+              .getAsJsonObject("topLevelComment")
+              .getAsJsonObject("snippet")
+              .get("likeCount")
+              .getAsInt();
+
+      String filteredComment = filterComment(topComment);
+      Comment commentAndLikes = new Comment(filteredComment, likes);
```
jessicaslaughter, line 212: nit: rename `commentAndLikes` to `comment`

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -213,10 +223,12 @@ private String getVideoId(String response) {
    * @param comments The array to be condensed.
    * @return A properly formatted String
    */
-  private String convertToString(ArrayList<String> comments) {
+  private String convertToString(ArrayList<Comment> comments) {
     StringBuilder res = new StringBuilder();
 
-    for (String comment : comments) {
+    for (Comment commentAndLikes : comments) {
+      String comment = commentAndLikes.text;
```
jessicaslaughter, line 230: nit: rename `commentAndLikes` to `comment`, and the current `comment` can be renamed to `commentText`

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -231,4 +243,24 @@ private String convertToString(ArrayList<String> comments) {
...
+  /**
+   * Filters out a comment string so that it's readable and doesn't break the frontend.
+   *
+   * @param comment The comment to be filtered.
+   * @return A properly formatted comment
+   */
+  private String filterComment(String comment) {
+    String filteredComment = comment;
+
+    // Handles non-ASCII characters, newlines,
```
jessicaslaughter, line 256: This comment should be updated now that you're not filtering out all non-ASCII characters

server/src/main/java/com/google/musicanalysis/types/Comment.java
```diff
@@ -3,12 +3,12 @@
...
-public class CommentLikes implements Serializable {
-  public String comment;
+public class Comment implements Serializable {
+  public String text;
   public Integer likes;
 
-  public CommentLikes(String comment, Integer likes) {
-    this.comment = comment;
+  public Comment(String text, Integer likes) {
+    this.text = comment;
```
jessicaslaughter, line 11: `comment` should be updated to `text`

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -231,4 +243,24 @@ private String convertToString(ArrayList<String> comments) {
...
+    // trimmed quotes, embedded escaped quotes.
+    // Order matters.
```
geekster777, line 258: This is one of those places where you want to do a thorough breakdown of what each part of this code does. Expect to have more comment than code. Something like:
```
// Strip away the following characters
// \u00a9 : copyright character
// \ufoo: foo character
// ...
filteredComment = filteredComment.replaceAll("(\u00a9|\u00ae|[\u2000-\u3300]|\ud83c[\ud000-\udfff]|\ud83d[\ud000-\udfff]|\ud83e[\ud000-\udfff])", "");
// Each comment should be on its own line so they are parsed as one large sentence
filteredComment = String.join("\n", filteredComment.split("\\\\n"));
...
```

server/src/main/java/com/google/musicanalysis/types/Comment.java
seunomonije: Sloppy by me lol thank you

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -213,22 +223,52 @@ private String getVideoId(String response) {
...
+   * @param comment The comment to be filtered.
+   * @return A properly formatted comment
+   */
+  private String filterComment(String comment) {
+    String filteredComment = comment;
+
+    // Removes 4683 different emojis and symbols so that NL and Perspective don't crash
+    // \u00a9 : copyright character
+    // \u00ae : registered sign
+    // \u2000-\u3300 : char code ranges 8192 to 13056
```
geekster777, line 259: This comment doesn't tell me anything that the code itself doesn't. What does this range represent?

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -213,22 +223,63 @@ private String getVideoId(String response) {
...
+    //     superscripts and subscripts, currency symbols, combining datacritical marks for symbols,
+    //     letterlike symbols, number forms, arrows, mathematical operators, miscellaneous technical,
+    //     control pictures, optical character recognition, enclosed alphanumerics, 
+    //     box drawing, block elements, geometric shapes, miscellaneous symbols,
+    //     dingbats, braille patterns, glagolitic, coptic, georgian supplement,
+    //     and more. Please see:
+    //     https://utf8-chartable.de/unicode-utf8-table.pl
+    //     for all the charcters in this range
+    //                         
+    //
```
geekster777, line 270: This is more verbose than necessary. Just say:
```
// \u2000-\u3300: superscripts, subscripts, and other symbols we can ignore.
```

PR #72, opened 2020-07-28
added weighted analysis of comments

seunomonije:
Also tweaked the limits to account for it


server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -95,6 +101,32 @@ private NLPResult analyzeWithNLP(String text) throws IOException {
...
+      for (Map.Entry<Integer, NLPResult> likesAndNLPResult : unweightedNLPMap.entrySet()) {
+          totalLikes += (double) likesAndNLPResult.getKey();
+      }
+
+      double weightedMagnitude = 0;
+      double weightedScore = 0;
+
+      for (Map.Entry<Integer, NLPResult> likesAndNLPResult : unweightedNLPMap.entrySet()) {
+          weightedMagnitude += (likesAndNLPResult.getKey()/totalLikes)*likesAndNLPResult.getValue().magnitude;
+          weightedScore += (likesAndNLPResult.getKey()/totalLikes)*likesAndNLPResult.getValue().score;
```
laptou, line 124: nit: put spaces around operators

client/js/analysis.js
```diff
@@ -130,9 +130,9 @@ function determineOverall(json, time) {
     isClear = 'CLEARLY ';
   }
 
-  if (score < -0.2) {
+  if (score < -0.35) {
     tone = 'NEGATIVE';
-  } else if (score > 0.2) {
+  } else if (score > 0.35) {
```
gracetian6, line 135: how are you getting these bounds?

seunomonije: https://cloud.google.com/natural-language/docs/basics#interpreting_sentiment_analysis_values

laptou: Please add that in a comment, I was wondering too

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -22,7 +23,7 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
       throws ServletException, IOException {
 
     String input = req.getParameter("name");
-
+
```
geekster777, line 26: nit: remove whitespace indentation here (it should just be newlines, no spaces)

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -57,11 +58,16 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
     NameAndChannel videoInfo = getNameAndChannel(nameJson);
 
     HashMap<String, String> perspectiveMap = analyzeWithPerspective(cumulativeComments);
-    NLPResult commentsSentiment = analyzeWithNLP(cumulativeComments);
+    
+    HashMap<Integer, NLPResult> unweightedNLPMap = new HashMap<>();
+    for (CommentLikes commentAndLikes: commentArray) {
+        unweightedNLPMap.put(commentAndLikes.likes, analyzeWithNLP(commentAndLikes.comment));
```
geekster777, line 64: Your total number of likes won't be unique. Instead make your key the NLPResult and your key's value the total number of likes. Otherwise you'll likely get collisions from comments with the same number of likes (which includes many comments with 1 or 0 likes).

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -95,6 +101,32 @@ private NLPResult analyzeWithNLP(String text) throws IOException {
...
+
+      for (Map.Entry<Integer, NLPResult> likesAndNLPResult : unweightedNLPMap.entrySet()) {
+          totalLikes += (double) likesAndNLPResult.getKey();
+      }
+
+      double weightedMagnitude = 0;
+      double weightedScore = 0;
+
+      for (Map.Entry<Integer, NLPResult> likesAndNLPResult : unweightedNLPMap.entrySet()) {
+          weightedMagnitude += (likesAndNLPResult.getKey() / totalLikes) * likesAndNLPResult.getValue().magnitude;
```
geekster777, line 123: Store the total number of likes in a variable instead of calling `likesAndNLPResult.getKey()`. This will improve code clarity.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -95,6 +101,32 @@ private NLPResult analyzeWithNLP(String text) throws IOException {
...
+   * out of the sum of the likes of the comments
+   *
+   * @param unweightedNLPMap the unweighted map of <like, nlp> pairs
+   * @return A weighted NLPResult object
+   */
+  private NLPResult createWeightedSentiment(HashMap<Integer, NLPResult> unweightedNLPMap) {
+      double totalLikes = 0;
+
+      for (Map.Entry<Integer, NLPResult> likesAndNLPResult : unweightedNLPMap.entrySet()) {
+          totalLikes += (double) likesAndNLPResult.getKey();
```
geekster777, line 116: Beware of dividing by 0 if there are no liked comments.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -95,6 +101,32 @@ private NLPResult analyzeWithNLP(String text) throws IOException {
...
+      }
+
+      double weightedMagnitude = 0;
+      double weightedScore = 0;
+
+      for (Map.Entry<Integer, NLPResult> likesAndNLPResult : unweightedNLPMap.entrySet()) {
+          weightedMagnitude += (likesAndNLPResult.getKey() / totalLikes) * likesAndNLPResult.getValue().magnitude;
+          weightedScore += (likesAndNLPResult.getKey() / totalLikes) * likesAndNLPResult.getValue().score;
```
geekster777, line 124: Shouldn't you just weigh magnitude? Weighing the score means that you change the meaning of the sentiment based on number of comments, which seems incorrect.

seunomonije: The score is how negative/positive a comment is, while the magnitude is how confident NLP is that it's correct basically. 

Here i'm saying that, to determine how negative/postiive the entire group is, I'm taking how negative/positive each comment is and proportionally adding it to a total magnitude and score value that reflects both the negative/positive-ness of a group as well as the confidence that group is negative/positive

geekster777: Ah, then we should only be modifying the score and not the magnitude then.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
```diff
@@ -95,6 +101,48 @@ private NLPResult analyzeWithNLP(String text) throws IOException {
...
+      for (Map.Entry<NLPResult, Integer> likesAndNLPResult : unweightedNLPMap.entrySet()) {
+          totalLikes += (double) likesAndNLPResult.getValue();
+      }
+      
+      if (totalLikes == 0) {
+          // If there are no likes on any comment, we they all need
+          // to be weighted equally. There's not enough data to confidently
+          // determine a communities reaction. Setting totallikes++ to avoid error.
+          // This will rarely hit.
+          totalLikes++;
```
geekster777, line 124: return a zeroed  NLPResult instead of changing what totalLikes actually means. Setting totalLikes to 1 when the value it represents is actually 0 is confusing and makes your code brittle.

server/src/main/java/com/google/musicanalysis/site/AnalysisServlet.java
seunomonije: My only case is how will I figure out the correct magnitude for each of the 20 comments then? If a comment which has a high number of likes has a high magnitude or "correctness", shouldn't that "correctness" also be weighted?

For example, if there's a comment that is is clearly positive, and gets 100k likes, shouldn't that clearness be taken into account more than a comment that is maybe positive and only gets 10k likes?

geekster777: Spoke offline. The magnitude should be weighted since we're trying to determine how confident we are on our overall score, so we want to give higher weight to our confidence in our scores that count for more.

PR #73, opened 2020-07-28
Heat map w/ hardcoded data

gracetian6:
Displays hard coded data that out of **13 total** liked videos, the **1st, 4th, and 7th latest videos are music**. (starting from index 0)

- used heat map in apex charts library to execute this
- data is hard coded as [1, 4, 7] and turned into a indicator array of size 13

Why a heat map??
- A user can easily tell how often they have liked music youtube videos by seeing how colorful the heatmap is. 

image


client/js/heat.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+// data will be from backend next time
+const TOTAL_LIKED = 13;
+// element value of x means xth latest video is music
+const LIKED_MUSIC_HIST = [1, 4, 7];
+
+// ith value of 1 means ith latest video is music
+const likedMusicBinaryHist = [...Array(TOTAL_LIKED)].fill(0);
```
laptou, line 17: That's a spicy way to create an array filled with zeros. Maybe add a comment that explains that this is what you are doing?

client/js/heat.js
```diff
@@ -0,0 +1,82 @@
...
+for (let i = 0; i < LIKED_MUSIC_HIST.length; i++) {
+  const el = LIKED_MUSIC_HIST[i];
+  likedMusicBinaryHist[el] = 1;
+}
+
+
+/**
+ * turns array of data into heat map
+ * @param {int[]} data array of all 0s/1s heat map data
+ * @param {int} dataLength size of data array
```
laptou, line 27: nit: `int` is not a JS type; in JS it's just called `number`

client/js/heat.js
```diff
@@ -0,0 +1,82 @@
...
+ * @returns {HeatRow[]} square matrix of HeatMapValues for HeatMap
+ */
+function createHeatMapValues(data, dataLength) {
+  // heat map should have equal length and width
+  const numRows = Math.ceil(Math.sqrt(dataLength));
+
+  const HeatMapValues = [];
+  let row;
+  for (let i = 0; i < dataLength; i = i + numRows) {
```
laptou, line 36: nit: `i += numRows` is easier to read

client/js/heat.js
```diff
@@ -0,0 +1,82 @@
...
+    row = new HeatRow(data.slice(i, i + numRows));
+    HeatMapValues.push(row);
+  }
+  return HeatMapValues.reverse();
+}
+
+/**
+ * renders heat map onto html
+ * @param {int[]} data array of all 0s/1s heat map data
+ * @param {int} dataLength size of data array
```
laptou, line 46: same thing with the `int`

client/js/heat.js
```diff
@@ -0,0 +1,82 @@
...
+ * turns array of data into heat map
+ * @param {int[]} data array of all 0s/1s heat map data
+ * @param {int} dataLength size of data array
+ * @returns {HeatRow[]} square matrix of HeatMapValues for HeatMap
+ */
+function createHeatMapValues(data, dataLength) {
+  // heat map should have equal length and width
+  const numRows = Math.ceil(Math.sqrt(dataLength));
+
+  const HeatMapValues = [];
```
laptou, line 34: local variables should be `camelCase`

client/js/heat.js
```diff
@@ -0,0 +1,82 @@
+class HeatRow {
```
laptou, line 1: maybe call this `HeatMapRow` instead? `HeatRow` sounds awkward

laptou (changes requested):

seunomonije (approved):

client/js/heat.js
```diff
@@ -0,0 +1,94 @@
...
+ * @returns binary arr of size total where ith value 1
+ * corresponds to value i in input arr
+ */
+function makeBinaryArr(arr, total) {
+  // first fill with 0s then 1s
+  const binaryArr = [...Array(total)].fill(0);
+  for (let i = 0; i < arr.length; i++) {
+    const el = arr[i];
+    binaryArr[el] = 1;
+  }
```
laptou, line 32: This loop would be clearer like this:
```javascript
for (const hotIndex of arr) {
  binaryArr[hotIndex] = 1;
}
```

client/js/heat.js
```diff
@@ -0,0 +1,94 @@
...
+function createHeatMapValues(data, dataLength) {
+  // heat map should have equal length and width
+  const numRows = Math.ceil(Math.sqrt(dataLength));
+
+  const heatMapValues = [];
+  let row;
+  for (let i = 0; i < dataLength; i += numRows) {
+    row = new HeatMapRow(data.slice(i, i + numRows));
+    heatMapValues.push(row);
+  }
```
laptou, line 51: Why did you declare `row` outside of the loop if you don't use it anywhere else? You could just `heatMapValues.push(new HeatMapRow(...))`

laptou (changes requested):

laptou (approved):

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+  /**
+   * format for a row for heat map values obj
+   * @param {number[]} data The array of 0s and 1s for a row in a heat map
+   */
+  constructor(data) {
+    this.data = data;
+  }
+}
+
+// data will be from backend next time
```
jessicaslaughter, line 11: nit: since you know this is something that needs to change in the future, you should use an official TODO comment like

`// TODO: retrieve data from the backend instead of using a hardcoded value`

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * converts an array to binary arr
+ * @param {number[]} arr values will be indices that hold 1 in output arr
+ * @param {number[]} total size of output arr
+ * @returns binary arr of size total where ith value 1
+ * corresponds to value i in input arr
+ */
+function makeBinaryArr(arr, total) {
+  // first fill with 0s then 1s
```
jessicaslaughter, line 27: nit: this comment isn't needed, it just explains the implementation below rather than the code's intentions. the code is easy enough to follow without further explanation.

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+ */
+function createHeatMapValues(data, dataLength) {
+  // heat map should have equal length and width
+  const numRows = Math.ceil(Math.sqrt(dataLength));
+
+  const heatMapValues = [];
+  for (let i = 0; i < dataLength; i += numRows) {
+    heatMapValues.push(new HeatMapRow(data.slice(i, i + numRows)));
+  }
+  return heatMapValues.reverse();
```
jessicaslaughter, line 49: add a comment explaining why you need to reverse the final result

jessicaslaughter (changes requested):

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+   * @param {number[]} data The array of 0s and 1s for a row in a heat map
+   */
+  constructor(data) {
+    this.data = data;
+  }
+}
+
+// data will be from backend next time
+const TOTAL_LIKED = 13;
+// element value of x means xth latest video is music
```
geekster777, line 13: nit: using terms like xth are difficult to follow. Say something closer to "list of day indices where music was played"

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+const LIKED_MUSIC_HIST = [1, 4, 7];
+
+// ith value of 1 means ith latest video is music
```
geekster777, line 16: This comment is difficult to parse. It sounds like this is an array of binary values (versus a binary tree being stored in an array, which is what I expect a "binary array" to be). If this is the case, store booleans instead where you can be more clear:
```
const likedMusicHist = markMusicDays(LIKED_MUSIC_HIST, TOTAL_LIKED);
```

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {number[]} arr values will be indices that hold 1 in output arr
+ * @param {number[]} total size of output arr
+ * @returns binary arr of size total where ith value 1
+ * corresponds to value i in input arr
+ */
+function makeBinaryArr(arr, total) {
+  // first fill with 0s then 1s
+  const binaryArr = [...Array(total)].fill(0);
+  for (const heatIndex of arr) {
+    binaryArr[heatIndex] = 1;
```
geekster777, line 30: as mentioned above, just set this value to `true` instead of giving it a number.

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+  return binaryArr;
+}
+
+/**
+ * turns array of data into heat map
+ * @param {number[]} data array of all 0s/1s heat map data
```
geekster777, line 37: This describes the type but not what "data" is. Instead elaborate that data is the list of days in which music was listened to.

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {number} dataLength size of data array
```
geekster777, line 38: This is unnecessary since you can just get this info from `data.length`

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+ * @returns {HeatMapRow[]} square matrix of HeatMapValues for HeatMap
+ */
+function createHeatMapValues(data, dataLength) {
+  // heat map should have equal length and width
```
geekster777, line 42: Good declarative comment.

jpnickolas (changes requested):

client/js/heat.js
gracetian6: I don't want to store booleans because the chart can only render integer values (what I have) not coolean values

client/js/heat.js
gracetian6: this is only cause dataLength is a hard coded value from the backend

client/js/heat.js
geekster777: Then leave a comment detailing this. Otherwise it's odd that you're representing a boolean values as a list of numbers. Something like:
```
// The histogram library takes a list of numbers so that heavier music
// days are displayed as darker. We don't want this gradient, so a day with
// any amount of music will be stored as a `1`.
```

client/js/heat.js
geekster777: It's still unnecessary. Ignore the `dataLength` you get from the back end and use `data.length`.

client/js/heat.js
geekster777: Resolving per previous comment thread.

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+
+// ith value of 1 means ith latest video is music
+const likedMusicBinaryHist = makeBinaryArr(LIKED_MUSIC_HIST, TOTAL_LIKED);
+
+/**
+ * converts an array to binary arr
+ * @param {number[]} arr values will be indices that hold 1 in output arr
+ * @param {number[]} total size of output arr
+ * @returns binary arr of size total where ith value 1
+ * corresponds to value i in input arr
```
geekster777, line 24: This comment is hard to follow and is essentially just verbalizing the code rather than explaining it. Say something like `Array of 1's and 0's where all indices provided in "arr" are set to 1`

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+const TOTAL_LIKED = 13;
+// list of day indices where music was played
+const LIKED_MUSIC_HIST = [1, 4, 7];
+
+// ith value of 1 means ith latest video is music
+const likedMusicBinaryHist = makeBinaryArr(LIKED_MUSIC_HIST, TOTAL_LIKED);
+
+/**
+ * converts an array to binary arr
+ * @param {number[]} arr values will be indices that hold 1 in output arr
```
geekster777, line 21: make this variable more descriptive. Something like `indicesToFlag`

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {number[]} total size of output arr
+ * @returns binary arr of size total where ith value 1
+ * corresponds to value i in input arr
+ */
+function makeBinaryArr(arr, total) {
```
geekster777, line 26: This function name is vague and just tells us about our data type. Given that you already use terms like `heatIndex` in your function body, you can make the name more specific (like `constructHeatMapSource`) even though the function itself is more general than this.

client/js/heat.js
```diff
@@ -0,0 +1,91 @@
...
+   * format for a row for heat map values obj
+   * @param {number[]} data binary arr of days where music was listened to
+   */
+  constructor(data) {
+    this.data = data;
+  }
+}
+
+// TODO: retrieve data from the backend instead of using a hardcoded value
+const TOTAL_LIKED = 13;
```
geekster777, line 12: nit: name this TOTAL_LIKED_VIDEOS

jpnickolas (changes requested):

gracetian6:
Thanks for your comments all! I don't have time to implement them since the internship is over.

PR #74, opened 2020-07-30
Vanilla bars w/ BACKEND DATA!!

gracetian6:
- integrate backend with front end bar chart
- uses ES modules for variable transfer
- fetched `GENRE_ANALYSIS` in `genre.js` and moved this JSON obj to `bar.js`

Backend data looks like: 
image

**FRONT END BAR CHART:**
image


client/js/bar.js
```diff
@@ -59,4 +54,9 @@ function createBarChart(chartValues, chartCategories) {
   }
 }
 
-createBarChart(CHART_VALUES, CHART_CATEGORIES);
+GENRE_ANALYSIS.then((DATA) => {
+  createBarChart(
+      Object.values(DATA.genreData),
+      Object.keys(DATA.genreData),
+      DATA.maxGenreCount);
```
gracetian6: I'm not sure if `Object.values` and `Object.keys` is the best way to do it

I assume that the array returned from both methods should match up with each other

client/.eslintrc
```diff
@@ -4,10 +4,12 @@
     "google"
   ],
   "env": {
-    "browser": true
```
gracetian6: this file is git pulled from master

client/js/util.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Delays an async function.
+ *
+ * @param {number} ms The number of milliseconds to wait before resolving the
```
gracetian6: this file is git pulled from master

client/js/bar.js
```diff
@@ -59,4 +54,9 @@ function createBarChart(chartValues, chartCategories) {
   }
 }
 
-createBarChart(CHART_VALUES, CHART_CATEGORIES);
+GENRE_ANALYSIS.then((DATA) => {
```
seunomonije: Don't think `DATA` is the best variable name here. Since you're getting a object, something like `genreAnalysisInfo` or something like that

client/js/bar.js
seunomonije: If they match up then I think this is a clever-ish solution to your problem. You could also just change createBarChart to accept the entire map and then loop through the key-value pairs there. I personally prefer the second method, but I'd also like to see what @laptou has to say about it.

client/js/util.js
seunomonije: If it's already in master why add it to this pull request?

client/.eslintrc
seunomonije: If it's already in master, why add it to this pull request?

seunomonije (changes requested):

client/js/util.js
gracetian6: I needed it in my local files so I can see the results.

It's not adding extra files, I just pulled it from master.

client/js/bar.js
laptou: +1; this is a parameter of a function and not a constant, so it should be in `camelCase`

client/js/genre.js
```diff
@@ -1,24 +1,24 @@
 /**
- * fetches youtube api calls from YoutubeServlet.java
+ * @file fetches youtube api calls from YoutubeServlet.java
  * and displays on youtube-genre.html
  */
 
-const genreBlock = document.getElementById('genres');
+document.body.onload = fetchMusicGenre();
```
laptou: Why are you calling `fetchMusicGenre()` as an onload handler if you call it on line 24?

client/js/genre.js
```diff
@@ -1,24 +1,24 @@
...
 
 /**
- * fetches genre count hashmap from /api/youtube and updates html
+ * fetches and returns genre analysis object from /api/youtube
+ *
+ * @returns {JSON object} of youtube genreData and stats
```
laptou: `JSON object` is not an ESLint type; if you don't want to write a specific typedef then you can just write `{object} An object containing YouTube genre data and stats`

client/js/util.js
laptou: yeah, it's just showing up in this PR b/c you didn't merge master into the branch this PR is based on. you should probably just do that, though.

laptou (changes requested):

laptou:
nit: there is no such thing as an ESLint module. these files are ES modules. ESLint is just a program that looks at your code and flags any stylistic errors.

client/js/util.js
seunomonije: ^

gracetian6: I will git pull everything from master

client/js/genre.js
gracetian6: O good point

seunomonije (approved):
Looks good!

client/js/bar.js
```diff
@@ -59,4 +54,9 @@ function createBarChart(chartValues, chartCategories) {
   }
 }
 
-createBarChart(CHART_VALUES, CHART_CATEGORIES);
+GENRE_ANALYSIS.then((genreAnalysisInfo) => {
```
laptou: I mentioned this in another review, but consider renaming `GENRE_ANALYSIS` to `GENRE_ANALYSIS_TASK` or `GENRE_ANALYSIS_PROMISE` so that it is extra clear that this object is a `Promise`.

laptou (approved):

client/js/bar.js
```diff
@@ -59,4 +54,9 @@ function createBarChart(chartValues, chartCategories) {
   }
 }
 
-createBarChart(CHART_VALUES, CHART_CATEGORIES);
+GENRE_ANALYSIS_PROMISE.then((genreAnalysisInfo) => {
+  createBarChart(
+      Object.values(genreAnalysisInfo.genreData),
```
jessicaslaughter, line 59: My opinion here would be to pass the entire map rather than separating it out. In createBarChart, we start to lose the connection between these two arrays when you're accessing values in one array based on the length of the other array. Open to other thoughts though @geekster777?

client/js/genre.js
```diff
@@ -1,24 +1,22 @@
...
     // no oauth login so redirect to new page
     window.open('/api/oauth/login/youtube');
   }
 
   const genreCount = await response.text();
-  genreBlock.innerHTML = genreCount;
+  return JSON.parse(genreCount);
 }
 
+export const GENRE_ANALYSIS_PROMISE = fetchMusicGenre();
```
jpnickolas: You shouldn't export a call to a promise. This line will trigger this fetch /anytime/ this file is imported. This may be what you want in the short term, but isn't what folks will expect from a library/file they import. You should instead export `fetchMusicGenre`, and call it in your other file.

client/js/bar.js
```diff
@@ -2,49 +2,44 @@
...
 const BAR_PERCENT_PADDING = 1 - BAR_PERCENT_FILL;
 
-const GRAPH_LEFT_PADDING = 2.5;
-
 /**
  * creates SVG bar chart given chart values and categories
+ *
  * @param {number[]} chartValues array of bar values/lengths
  * @param {string[]} chartCategories array of bar categories/labels
+ * @param {number} maxChartVal largest value of chartValues
```
jpnickolas: You shouldn't pass this value in. Instead compute it from `chartValues`.

client/js/bar.js
jpnickolas: +1. It makes your function call more concise, and like Jess said, keeps the direct connection between your categories and their values.

client/js/bar.js
```diff
@@ -59,4 +54,9 @@ function createBarChart(chartValues, chartCategories) {
   }
 }
 
-createBarChart(CHART_VALUES, CHART_CATEGORIES);
+GENRE_ANALYSIS_PROMISE.then((genreAnalysisInfo) => {
```
jpnickolas: As stated below, just import fetchMusicGenre and call it here.

client/js/genre.js
jpnickolas: In general, libraries and imported files shouldn't have side effects. They should be idempotent.

jpnickolas (changes requested):

client/js/genre.js
gracetian6: oh that is right

PR #75, opened 2020-07-30
Backend for Heat Map

gracetian6:


client/css/bar.css
```diff
@@ -0,0 +1,60 @@
...
+  align-content: center;
+  justify-items: center;
+  width: 100%; /* don't use 100vw, that will cause horizontal scrollbar */
+  height: 100vh;
+}
+
+.svg {
+  width: 100%;
+  height: 100%;
+}
```
gracetian6, line 12: ignore changes on bar.css and bar.js .... I don't know why they are appearing

server/src/main/java/com/google/musicanalysis/site/YoutubeGenres.java
```diff
@@ -68,12 +74,17 @@ protected void calculateMusicCount(JsonArray videos) {
             }
         }
 
-        if (isMusic && totalSubgenres == 0) {
+        if (isMusic) {
+          // likedMusicHistory records video numbers that are music
+          likedMusicHistory.add(firstVideoCount + i);
```
seunomonije, line 79: Took me a minute to see what you're doing, but now it makes sense. One thing that I noticed tho is that you're passing in a 0 for `firstVideoCount` in your `YoutubeServlet.java` file, which to me makes it useless since you're just adding by zero every time. Why are you passing it through this function? Will there be a case where you pass in a number other than 0?

If you want to keep track of the number of videos retrieved, you could just do
 ```
int videosRetrieved = 0;
...
videosRetrieved += genreAnalysis.calculateMusicCount(videos);
```
since it looks like you're just returning the size of the `videos` jsonArray. 

You also never modify the `videos` jsonArray, so what's the point of returning the size with this function? I think leaving this as a `void` function was more clear.

File: server/src/main/java/com/google/musicanalysis/site/YoutubeGenres.java [hunk 3]
```diff
@@ -68,12 +74,17 @@ protected void calculateMusicCount(JsonArray videos) {
...
+
+          if (totalSubgenres == 0) {
             // video only classified as Music so we update as "Other music"
-            this.updateGenre("Other music");
+            this.updateGenre("Other music");            
+          }
         }
     }
-    return;
+    return firstVideoCount + videos.size();
```
seunomonije, line 87: nit: Rather than calling `videos.size()` twice, it might be more efficient to store that in a variable and use it throughout. I think technically this is okay because `.size()` is constant? (not sure) but if it takes O(n) and you're not modifying the array there's no need to call this method twice

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -111,7 +113,8 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
             }
 
             videos = likedVideoRes.getAsJsonArray("items");
-            genreAnalysis.calculateMusicCount(videos);
+            videosRetrieved = genreAnalysis
```
seunomonije, line 116: I think this section can be clearer. See above

client/css/bar.css
laptou: is this PR supposed to be based against `master`?

gracetian6: I will git pull everything from master :'(

server/src/main/java/com/google/musicanalysis/site/YoutubeGenres.java
gracetian6: got it, I'll make these changes

client/js/genre.js
```diff
@@ -3,12 +3,13 @@
...
-const genreBlock = document.getElementById('genres');
 
 /**
  * fetches genre count hashmap from /api/youtube and updates html
  */
 async function displayMusicGenre() {
+  const genreBlock = document.getElementById('genres');
+
   // keep track of num_videos in URL w/o reload
   history.pushState('', '', `youtube-genre.html`);
```
laptou, line 14: How does this keep track of `num_videos`? Sorry I missed this in an earlier PR.

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -96,9 +96,13 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
         JsonObject likedVideoRes;
         JsonArray videos;
         YoutubeGenres genreAnalysis = new YoutubeGenres();
+        // needed to keep track of likedMusicHistory for heat map
```
jessicaslaughter, line 99: this comment doesn't really make sense to me, since it's not "keeping track" of `likedMusicHistory` but being used by it. `likedMusicHistory` also isn't part of this file, so that could come of as confusing to someone only looking at this file. consider rephrasing to something like "used to determine the position of a newly retrieved video in comparison to already retrieved videos while obtaining more videos in segments"

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -111,7 +115,9 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
             }
 
             videos = likedVideoRes.getAsJsonArray("items");
-            genreAnalysis.calculateMusicCount(videos);
+            // videosRetrieved keeps track of music video order in genreAnalysis
```
jessicaslaughter, line 118: You don't need this comment since you already have a comment defining what this variable is when you initiated it

server/src/main/java/com/google/musicanalysis/site/YoutubeGenres.java
```diff
@@ -31,9 +35,12 @@ public YoutubeGenres() {
  * parses through youtube liked videos json array,
  * updates hash map to contain frequency count of each music genre
  * @param videos json array of youtube liked videos
+ * @param firstVideoCount the number of videos retrieved before this http call
```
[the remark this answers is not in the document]
geekster777: +1. Probably something like "currentVideoIndex" will be clearest

server/src/main/java/com/google/musicanalysis/site/YoutubeGenres.java
```diff
@@ -68,12 +75,17 @@ protected void calculateMusicCount(JsonArray videos) {
...
+
+          if (totalSubgenres == 0) {
             // video only classified as Music so we update as "Other music"
-            this.updateGenre("Other music");
+            this.updateGenre("Other music");            
+          }
         }
     }
-    return;
+    return videosSize;
```
geekster777, line 88: You can get this info from the parent function. No need to return it here, and that in fact obscures what this function is doing, since it analyzes the videos but also returns the total number. Someone interacting with this function will likely write:
```
totalVideos += calculateMusicVideos(videos, totalVideos);
```

This means 2 independent things are happening on one line that aren't strictly tied together. More clear is:
```
calculateMusicVideos(videos, totalVideos);
totalVideos += videos.size();
```

server/src/main/java/com/google/musicanalysis/site/YoutubeServlet.java
```diff
@@ -111,7 +115,9 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res)
...
+            videosRetrieved += genreAnalysis
+                                .calculateMusicCount(videos, videosRetrieved);
```
geekster777, line 120: See comment above. This should be split into 2 separate lines.

PR #77, opened 2020-07-31
Heat Map with Backend Data

gracetian6:


client/js/genre.js
```diff
@@ -1,24 +1,23 @@
...
     // no oauth login so redirect to new page
     window.open('/api/oauth/login/youtube');
   }
 
   const genreCount = await response.text();
-  genreBlock.innerHTML = genreCount;
+  return JSON.parse(genreCount);
 }
+
+export const GENRE_ANALYSIS = fetchMusicGenre();
```
laptou, line 23: consider renaming this to `GENRE_ANALYSIS_PROMISE` or `GENRE_ANALYSIS_TASK`; it needs to be very obvious that this object is a `Promise` for your code in the other files to make sense right away.

gracetian6: good idea!! 😀

client/js/heat.js
```diff
@@ -35,28 +37,26 @@ function makeBinaryArr(arr, total) {
...
     heatMapValues.push(new HeatMapRow(data.slice(i, i + numRows)));
   }
   return heatMapValues.reverse();
 }
 
 /**
  * renders heat map onto html
- * @param {number[]} data array of all 0s/1s heat map data
- * @param {number} dataLength size of data array
+ * @param {number[]} allBinaryData array of all 0s/1s heat map data
```
jessicaslaughter, line 55: consider renaming to just `binaryData`, `allBinaryData` insinuates that there's some other array with just a segment of the data, which there isn't so we can simplify this name

client/js/heat.js
```diff
@@ -88,4 +88,9 @@ function createHeatMap(data, dataLength) {
   chart.render();
 }
 
-createHeatMap(likedMusicBinaryHist, TOTAL_LIKED);
+GENRE_ANALYSIS.then((genreAnalysisInfo) => {
+  const LIKED_MUSIC_BINARY_HIST = makeBinaryArr(
```
jessicaslaughter, line 92: nit: I don't think this is a true constant, it should just use lowerCamelCase

client/js/genre.js
```diff
@@ -1,24 +1,23 @@
...
     // no oauth login so redirect to new page
     window.open('/api/oauth/login/youtube');
   }
 
   const genreCount = await response.text();
-  genreBlock.innerHTML = genreCount;
+  return JSON.parse(genreCount);
 }
+
+export const GENRE_ANALYSIS_PROMISE = fetchMusicGenre();
```
geekster777, line 23: See earlier PR about not calling a function with side effects (such as making a web request) on import

PR #79, opened 2020-07-31
GDPR Chart zoom (part 1)

laptou:
- Implement basics for GDPR chart zoom
- Add `update` functions to chart to adjust SVG elements when user zooms


client/js/gdpr/chart.js
```diff
@@ -290,10 +290,10 @@ export class GdprChart extends HTMLElement {
     endCap.setAttribute('class', 'series-run-cap');
 
     startCap.setAttribute('r', '5');
-    startCap.setAttribute('cx', points[0].x + 'px');
+    startCap.setAttribute('cx', points[0].x * this.zoom + 'px');
     startCap.setAttribute('cy', points[0].y + 'px');
     endCap.setAttribute('r', '5');
-    endCap.setAttribute('cx', points[points.length - 1].x + 'px');
+    endCap.setAttribute('cx', points[points.length - 1].x * this.zoom + 'px');
     endCap.setAttribute('cy', points[points.length - 1].y + 'px');
```
gracetian6, line 297: why is this.zoom only being update for cx, x1, x2 but not y positions?

gracetian6: also is this.zoom being updated anywhere? (I don't see it)

laptou: - the zoom only affects horizontal scaling, not vertical
- `this.zoom` is not updated anywhere, that part is added in #80

client/js/gdpr/chart.js
```diff
@@ -10,6 +10,8 @@ export class GdprChart extends HTMLElement {
 
     this.attachShadow({mode: 'open'});
 
+    this.zoom = 1;
+
```
gracetian6, line 14: add a comment here describing what this.zoom does, and how it only affects horizontal elements

client/js/gdpr/chart.js
```diff
@@ -275,7 +275,7 @@ export class GdprChart extends HTMLElement {
         }));
 
     const pointsStr = points
-        .map(({x, y}) => `${x}, ${y}`)
+        .map(({x, y}) => `${x * this.zoom}, ${y}`)
```
geekster777, line 278: You should apply your zoom earlier when you apply RUN_SCALE_X to the points. This will keep you from needing to re-apply zoom below on your start and end caps.

client/js/gdpr/chart.js
```diff
@@ -416,6 +416,91 @@ export class GdprChart extends HTMLElement {
...
+   * leaderboard.
+   * @param {number} start The index of the first entry in this run.
+   * @param {number} end The index of the last entry in this run.
+   * @private
+   */
+  updateRun(runContainer, history, start, end) {
+    const points = history
+        .slice(start, end)
+        .map((val, idx) => ({
+          x: (idx + start) * RUN_SCALE_X,
```
geekster777, line 472: Same. Apply the zoom here so your modifiers are all in one place.

PR #80, opened 2020-07-31
GDPR Chart Zoom (part 2)

laptou:
- Modify styles to accommodate zooming
- Connect zoom to mouse wheel event


client/js/gdpr/chart.js
```diff
@@ -114,8 +115,29 @@ export class GdprChart extends HTMLElement {
     this.clearHover();
   }
 
-  onScroll() {
-    this.clearHover();
+  onMouseWheel(ev) {
+    if (ev.ctrlKey) {
```
gracetian6, line 119: so if the cltr key is pressed then we update the zoom, otherwise its cleared?

client/js/gdpr/chart.js
```diff
@@ -424,6 +450,15 @@ export class GdprChart extends HTMLElement {
    * @private
    */
   updateChart() {
+    this.svg.setAttribute(
+        'viewBox',
+        '0 ' +
+      // shift down so that lines are vertically centered
+      RUN_SCALE_Y * 0.5 + ' ' +
+      this.dates.length * RUN_SCALE_X * this.zoom + ' ' +
+      NUM_POSITIONS * RUN_SCALE_Y,
```
gracetian6, line 459: indent this so it lines up with '0 ' + in line 455

client/js/gdpr/chart.js
```diff
@@ -114,8 +115,29 @@ export class GdprChart extends HTMLElement {
...
   }
 
-  onScroll() {
-    this.clearHover();
+  onMouseWheel(ev) {
+    if (ev.ctrlKey) {
+      const {min, max, pow} = Math;
+      const scale = pow(10, ev.deltaY / 100);
+
+      const finalZoom = min(max(this.zoom * scale, 0.1), 3);
```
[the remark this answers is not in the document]
laptou: Mouse wheel delta is a positive value or a negative value depending on which way the user is scrolling, but it's also linear. I need to use a logarithm in order for the scaling to feel linear when you scroll (i.e., scrolling one tick inwards should move you from 0.5x to 0.4x, but having the same movement scroll you from 1.5x to 1.6x would feel inconsistent). Finally, the min and max are so that the user can't scroll too far in either direction.

client/js/gdpr/chart.js
laptou: yes

client/js/gdpr/chart.js
```diff
@@ -117,8 +118,29 @@ export class GdprChart extends HTMLElement {
...
```
jessicaslaughter, line 126: You should add a brief comment explaining why this calculation is necessary

geekster777: +1, you've got several magic numbers and exponents that could use explaining.

client/js/gdpr/chart.js
```diff
@@ -117,8 +118,29 @@ export class GdprChart extends HTMLElement {
...
+      const finalScale = finalZoom / this.zoom;
+
+      // point under user's cursor should not move on scroll
```
geekster777, line 129: Good declarative comment

